PR #4, opened 2015-07-31
Support custom keyboard

### LINKED ISSUE #1: Reply keyboards
https://core.telegram.org/bots/api#replykeyboardmarkup

We still miss custom reply keyboards (which, so far, do work on mobiles only). Since I lost my cell phone, I am currently not being able to hack and, therefore, test these. If anyone needs this functionality, you are totally welcome to pull!

Ronmi:
As in [Official API document](https://core.telegram.org/bots/api/#sendmessage), custom keyboard and force reply are packed in same parameter. So I renamed `ForceReply` to `ReplyMarkup` to fit this.

This should fix #1

### DIFF
--- a/api.go
+++ b/api.go
@@ -95,9 +95,10 @@ func embedSendOptions(params *url.Values, options *SendOptions) {
 		params.Set("disable_web_page_preview", "true")
 	}
 
-	if options.ForceReply.Require {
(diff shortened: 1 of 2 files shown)


tucnak:
Seems legit, great work!

PR #5, opened 2015-08-04
Doing long polling in Bot.Listen to ensure message order

Ronmi:
This commit ensures we will not receive same message multiple times when in bad network-speed environment. And message will be put to channel in same order we receive from API call.

 * Bot.Listen(): Use only 1 goroutine. Do really long polling, not repeatly do short polling.
 * API.go: getUpdates(): as doing long polling, channel is not required.


tucnak:
And once again, awesome job!

PR #12, opened 2015-09-14
Support new API: markdown message.

Ronmi:


options.go
```diff
@@ -1,5 +1,14 @@
 package telebot
 
+// ParseMode is new bot api released at Sep 7, 2015. It adds markdown support to sendMessage API
```
tucnak: I'd better change it to smth like "ParseMode determines the way client applications treat the text of the message", makes much more sense to me (a lib user wouldn't really care about API history).

tucnak:
Looks great though. Nice to see that there is someone who actually keeps track on these API updates!

Ronmi:
Thanks for suggestion :)

PR #14, opened 2015-10-11
Change API to return error on 500 http status.

ejamesc:
- The Telegram Bot API does not return well-formed JSON when there is an internal server error. It instead returns HTML. This results in incomprehensible errors from attempting to unmarshal the HTML as JSON.
- This pull request changes behaviour to look at the http status from the HTTP Client's response before creating and returning an error. 

Note: The Telegram bot API seems to return proper JSON for every other status code, though.


tucnak:
Looks reasonable, thanks!

PR #15, opened 2015-10-13
use go routine in processing message channel

aladine:


tucnak:
Not really sure that it's the good way to go...

aladine:
The return result of getUpdate function will limit to 100 messages by default. So you never execute more than 100 go routines in parallel.

tucnak:
Alright, let's see.
1) Have you tried buffering the subscription list? E.g. `make(chan telebot.Message, 1000)` wouldn't block `getUpdates` feed until channel fills with 1000 messages.
2) Is wrapping the subscription range loop really does magic? If it does, we might indeed stick to this approach and outline it in docs carefully.

aladine:
Not sure yet, I will test this scheme when my bot is more stable. Now occasionally, it still encountered the error in another issue I informed.

ejamesc:
@aladine My approach is to handle each new command in a separate goroutine, because some commands I process take a bit of processing time (e.g. to fetch images, search results, and so on). So I totally agree with you that one should just spawn a goroutine to handle each command - in practice I haven't had any trouble with this.

However, I would caution against using this as an example, though. One problem is that if a panic is every encountered in any of the goroutines, your entire program will crash. 

The workaround is to wrap the goroutine with a function, like so: https://github.com/ejamesc/jarvisbot/blob/master/jarvisbot.go#L155

tucnak:
@ejamesc Serving responses in the distinct goroutines is obviously the way to go for "heavy" usecases, which you don't serve like instantly. For-range is a nice example, but I think we should provide some additional API for serving stuff (with e.g. command handling functionality), so the following code actually makes sense:

```go
bot, err := telebot.NewBot("SECRET TOKEN")
if err != nil {
	return
}

bot.Handle("/start", func(ctx telebot.Context) {
	// this function is executed per-message, in the distinct goroutine
})

bot.Serve(1 * time.Second())
```

olebedev:
`bot.Handle` LGTM, will you add?

tucnak:
@olebedev I certainly could, but I'm rather busy these days. Anyway, I'll try to get some time and finish it this week.

suntong:
Following up from the saying of 

> Also, consider Telebot a highload-ready solution.

in readme, 

> @ejamesc Serving responses in the distinct goroutines is obviously the way to go for "heavy" usecases, which you don't serve like instantly. For-range is a nice example, but I think we should provide some additional API for serving stuff (with e.g. command handling functionality), so the following code actually makes sense:
> 
> ```go
> bot, err := telebot.NewBot("SECRET TOKEN")
> if err != nil {
>     return
> }
> 
> bot.Handle("/start", func(ctx telebot.Context) {
>     // this function is executed per-message, in the distinct goroutine
> })
> 
> bot.Serve(1 * time.Second())
> ```

Has _"// this function is executed per-message, in the distinct goroutine"_ been implemented by the framework, or we need to spawn goroutines ourselves in bot.Handle call back functions?

tucnak:
@suntong Each and every handler is run as a goroutine.

PR #22, opened 2015-11-03
Fix README example

shoonoise:
Seems like example was outdated.


tucnak:
Indeed! Nice catch, danke!

PR #23, opened 2015-11-12
Fix close connection after usage.

reddec:
Close connection after usage.

Due to [offical documentation](https://golang.org/pkg/net/http/#Response), use `resp.Close = true` .


tucnak:
Thanks!

PR #24, opened 2015-11-12
add webHook support and push updates

reddec:
Adds push updates by webHook. It MUST be used with SSL certifications due to Telegram rules.

Could be used like this:

```go
err = bot.ListenWebHook(messages, 
	"<yourhost>:<your port>", 
	"/path/on/your/server", 
	"your_cert.pem", 
	"your_priv.key")
if err != nil {
	panic(err)
}
```

Address has to be always with port number

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -3,8 +3,10 @@ package telebot
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


tucnak:
I am still not sure that we do really need this in the project. Can you provide the use case of webhooks being superior to long polling?

bot.go
```diff
@@ -47,6 +49,41 @@ func (b Bot) Listen(subscription chan<- Message, timeout time.Duration) {
 	}()
 }
 
+// ListenWebHook starts HTTPS server and register webHooks in Telegram for PUSH
+// updates. certFile and keyFile - filenames containing a certificate and matching private key for the server
+// See https://golang.org/pkg/net/http/#ListenAndServeTLS
+func (b Bot) ListenWebHook(subscription chan<- Message, publicAddress, path, certFile, keyFile string) error {
```
tucnak: Way too clumsy, imo we should go with a distinct struct, probably embedding `http.Server`.

reddec:
1. No persistence connection - lower resource utilization.
2. Minimal delay
3. And it just may works out-of-box in Go)

bot.go
```diff
@@ -47,6 +49,41 @@ func (b Bot) Listen(subscription chan<- Message, timeout time.Duration) {
...
+	// register web hook in Telegram
+	if _, err := sendFile("setWebhook", b.Token, "certificate", certFile, url.Values{"url": []string{publicAddress + path}}); err != nil {
```
tucnak: You deffo shouldn't go with `sendFile` here, it's not about sending files at all :-1:

bot.go
reddec: Yes and no:
1. Future users must always remember, that SSL connection is required - using just http.Server is not requires this fact
2. Probably we may extend this feature in next PR if it really needs

bot.go
```diff
@@ -47,6 +49,41 @@ func (b Bot) Listen(subscription chan<- Message, timeout time.Duration) {
...
+			decoder := json.NewDecoder(request.Body)
+			err := decoder.Decode(&update)
+			if err != nil {
+				http.Error(response, err.Error(), 502)
+				return
+			}
+			subscription <- update.Payload
+			response.WriteHeader(200)
+		})
+		panic(server.ListenAndServeTLS(certFile, keyFile))
```
tucnak: I am not quite sure we should panic here, since it'll crash the whole app if not recovered.

bot.go
reddec: We have to `sendFile` due to uploading public certificate

bot.go
reddec: Yes, it is really not best solution, but if you have better ideas - let's try to improve together

bot.go
tucnak: We could go for a wrapper then. The current implementation is way too quirky for the API niceness level, telebot aims for.

reddec: Ok, what about returning `http.HandleFunc, error` as decision for both problems (error and universal api)?

PR #32, opened 2016-03-16
Add inline response options and offset handling

zenitraM:
I've added support for sending [Inline Query response options](https://core.telegram.org/bots/api#answerinlinequery), such as cache time, cache personality, and offset (which is now parsed from the response).

Note this changes the API to make it similar to the message sending functions, I haven't figured out a way to do it cleanly without that...

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ func queries() {
 		results := []telebot.Result{...}
 
 		// And finally respond to the query:
-		if err := bot.Respond(query, results); err != nil {
+		if err := bot.Respond(query, results, nil); err != nil {
 			log.Println("ouch:", err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


tucnak:
I totally can see how it's useful and practical, yet I'm all against ruining backwards compat.

P.S. Sorry for a postponed response :|

zenitraM:
Yes :disappointed: that is the part I don't like at all.

How would you prefer me to add this without breaking the API? maybe adding a new `RespondWithOptions` function?

tucnak:
I got a better idea. Let's define a `type Results []Result` with optional fields and use it instead of `[]Result` all over the place. This will introduce the functionality and won't break the API.

tucnak:
Any chance you are willing to implement the said behaviour?

zenitraM:
Sorry for the delay - I might not be able to get back to this until a couple weeks from now. Feel free to implement it if you can get to it sooner than I do :)

PR #35, opened 2016-04-03
Added basic routing features

rodrigodiez:
Hi,

I've been using this useful bot for a few days and I wanted to contribute with a basic routing feature. 

It analyses messages against regexps, finds the right controller to handle the message and builds a map of named arguments so the controller can use them

I am quite new at go developing so please point to any issue and I will be happy to solve it :)

``` go
// bot initialization

router := telebot.NewRouter()
router.When("/test (?P<first>[a-z]+) (?P<second>[a-z]+)", testController)

messages := make(chan telebot.Message)
for message := range messages {
    controller, args, err := router.GetController(&message)
    checkErr(err)

    err = controller(&message, args)
    checkErr(err)
}

// controller
func testController(message *telebot.Message, args *map[string]string) error {
    bot.SendMessage(message.Chat, fmt.Sprintf("%v", *args), nil)
    return nil
}
```


tucnak:
Hey there!

Thanks for contributing, I really appreciate it. The code looks practical, yet sort of "not idiomatic nuff". I'd like to see smth more consistent with Go stdlib and/or popular libraries. Nother nitpick: you shouldn't pass `map` by pointer, it's a reference type, effectively passed around by value.

Btw, have you seen PR #25? I had been working on it for a while when got all swamped with my studies.

rodrigodiez:
Thanks for your feedback! I will try to improve the code and merge some approaches you took on #25 like exposing the api directly through Bot and implementing Context. Stay tunned

rodrigodiez:
I got this working on my bot. Quite handy!

```go
bot.Handle("/test (?P<name>[a-z]+) (?P<last_name>[a-z]+)", func(context telebot.Context) {
  bot.SendMessage(context.Message.Chat, fmt.Sprintf("Hello %s, %s", context.Args["last_name"], context.Args["name"]), nil)
})

bot.Handle("", func(context telebot.Context) {
  bot.SendMessage(context.Message.Chat, "What???", nil)
})

bot.Serve()
```

rodrigodiez:
@tucnak did you had a look? I am eager for feedback :D

bot.go
```diff
@@ -5,16 +5,19 @@ import (
...
 type Bot struct {
-	Token    string
-	Identity User
-	Messages chan Message
-	Queries  chan Query
+	Token          string
+	Identity       User
+	Messages       chan Message
+	Queries        chan Query
+	Handlers       map[string]Handler
```
tucnak: I'd split the last two routing-related props out!

bot.go
```diff
@@ -5,16 +5,19 @@ import (
...
+	DefaultHandler Handler
```
tucnak: Why do we need it?

bot.go
```diff
@@ -28,6 +31,7 @@ func NewBot(token string) (*Bot, error) {
 	return &Bot{
 		Token:    token,
 		Identity: user,
+		Handlers: make(map[string]Handler),
```
tucnak: It's fine, but `map[string]Handler{}` syntax is generally more common.

bot.go
```diff
@@ -507,3 +511,45 @@ func (b *Bot) Respond(query Query, results []Result) error {
...
+			handler(Context{Message: &message, Args: args})
+		}
+	}
+}
+
+func (b *Bot) route(message *Message) (Handler, map[string]string, error) {
+	for i, handler := range b.Handlers {
+		r, err := regexp.Compile(i)
+		if err != nil {
+			return nil, nil, err
```
tucnak: Err is being returned and not getting handled properly at :529. If I were you I would have compiled all of the regular expressions in `Serve()` inb4 actually serving and used `MustCompile()` in order to get a straight panic (that's what you typically do with regexps.

context.go
```diff
@@ -0,0 +1,10 @@
+package telebot
+
+const Default string = ""
```
tucnak: There we go, default handler is here, why would you need to introduce `Bot.DefaultHandler` then?

rodrigodiez: I guess dont know whether the concept of a default handler itself is beneficial or is just overcomplicating things with no real value for the user. I just imported the idea form your PR without thinking too much about it and now it does not feel right

I think for now I will stick to an MVP approach. If route() can not find any valid handler, no handler is executed

rodrigodiez:
@tucnak thanks for your feedback! I've improved the code, dropped some unclear features, reorganised the files a bit and updated the readme

bot.go
```diff
@@ -507,3 +510,39 @@ func (b *Bot) Respond(query Query, results []Result) error {
...
+// Handle registers a handler for a message which text matches the provided regular expression
+func (b *Bot) Handle(command string, handler Handler) {
+	reg := regexp.MustCompile(command)
+	b.handlers[reg] = handler
+}
+
+// Serve listens for messages and route them to the appropiate handler
+func (b *Bot) Serve() {
+	messages := make(chan Message)
+	b.Listen(messages, 1*time.Second)
```
zabawaba99: I'm a bit new to the Telegram bot API. Is there a reason this is hard coded to 1 second? Do they do some sort of rate limiting? It would be nice if this would be configurable.

tucnak:
I can't believe I came back to it only now. Would you please mind resolving existing conflicts, so I could make it to master?

tucnak:
Closing in favour of #25, which incorporates the code from this PR with proper author attribution.

PR #37, opened 2016-04-07
send Audio extra params

grunmax:
add all params for Audio
// all of them are in multipart-data body, but can`t see any difference in reply


tucnak:
Are you sure about the branch?

grunmax:
Cherry-pick is inconvenied could be, I'll take branch with last code better.

tucnak:
What do you mean "last code"? Actual development branch is master, can you please rebase against it so I could pour your code in?

PR #50, opened 2016-06-26
Improve support for inline queries

zoni:
These changes make it possible to set custom options when responding to
inline queries, as described on https://core.telegram.org/bots/api#answerinlinequery.

It also includes all the (non-cached) inline result types as described at https://core.telegram.org/bots/api#inlinequeryresult.

Some remarks:
* The internals of sendCommand have changed. It now expects a
  JSON-serializable object. Instead of doing GET requests with
  URL-encoded query parameters it now POSTs JSON directly.
* Because of the above, sendFile() has changed as well. It now expects a
* `map[string]string` which it will internally convert to URL encoded
  form values.
* Respond has been deprecated in favor of the new AnswerInlineQuery
  function. It is only kept for backward compatibility.
* A dependency on https://github.com/mitchellh/hashstructure has been
  introduced in order to generate automatic IDs for inline results.

Closes #32.

### DIFF
--- a/api.go
+++ b/api.go
@@ -8,17 +8,20 @@ import (
 	"io/ioutil"
 	"mime/multipart"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
 )
 
-func sendCommand(method, token string, params url.Values) ([]byte, error) {
-	url := fmt.Sprintf("https://api.telegram.org/bot%s/%s?%s",
-		token, method, params.Encode())
+func sendCommand(method, token string, payload interface{}) ([]byte, error) {
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/%s", token, method)
 
-	resp, err := http.Get(url)
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(payload); err != nil {
+		return []byte{}, err
+	}
+
+	resp, err := http.Post(url, "application/json", &b)
 	if err != nil {
 		return []byte{}, err
 	}
@@ -32,7 +35,7 @@ func sendCommand(method, token string, params url.Values) ([]byte, error) {
 	return json, nil
 }
 
-func sendFile(method, token, name, path string, params url.Values) ([]byte, error) {
+func sendFile(method, token, name, path string, params map[string]string) ([]byte, error) {
 	file, err := os.Open(path)
 	if err != nil {
 		return []byte{}, err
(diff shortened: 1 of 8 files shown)


zoni:
If accepted, this closes #32.

types.go
```diff
@@ -138,6 +138,13 @@ type KeyboardButton struct {
 	InlineQuery string `json:"switch_inline_query,omitempty"`
 }
 
+// InlineKeyboardMarkup represents an inline keyboard that appears right next
+// to the message it belongs to.
+type InlineKeyboardMarkup struct {
+	// Array of button rows, each represented by an Array of KeyboardButton objects.
+	InlineKeyboard [][]KeyboardButton `json:"inline_keyboard,omitempty"`
```
zoni: The `KeyboardButton` type in fact implements [InlineKeyboardButton](https://core.telegram.org/bots/api#inlinekeyboardbutton).

[KeyboardButton](https://core.telegram.org/bots/api#keyboardbutton) is a different entity and not currently represented by anything in telebot. I'd love to correct this but unfortunately that would break the current API.

api.go
```diff
@@ -8,17 +8,20 @@ import (
...
 
-func sendCommand(method, token string, params url.Values) ([]byte, error) {
-	url := fmt.Sprintf("https://api.telegram.org/bot%s/%s?%s",
-		token, method, params.Encode())
+func sendCommand(method, token string, payload interface{}) ([]byte, error) {
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/%s", token, method)
+
+	var b bytes.Buffer
+	if err := json.NewEncoder(&b).Encode(payload); err != nil {
+		return []byte{}, err
```
thehowl: Slices can be nil. In this case, I'd say to just go with a `return nil, err`.

inline_types.go
```diff
@@ -0,0 +1,690 @@
...
+
+	return json.Marshal(struct {
+		wrappedInlineQueryResultContact
+		Type string `json:"type"`
+		ID   string `json:"id",hash:"ignore"`
+	}{
+		wrappedInlineQueryResultContact: wrappedInlineQueryResultContact(*r),
+		ID:   id,
+		Type: "contact",
+	})
```
thehowl: As you've just defined the struct and you're giving a value to all the struct's fields, wouldn't it be better to omit the field names? Like this:

```
	return json.Marshal(struct {
		wrappedInlineQueryResultArticle
		Type string `json:"type"`
		ID   string `json:"id",hash:"ignore"`
	}{wrappedInlineQueryResultArticle(*r), id, "article"})
```

This comment also applies to all the other inline types.

api.go
zoni: That's a good point, however `return []byte{}, err` was already used in the original code so I kept it consistent with that behavior.

inline_types.go
zoni: I don't really have a strong opinion one way or the other. I'll agree it does read slightly better.

tucnak:
Huge thanks for the great job done here, my appreciation!

PR #55, opened 2016-07-25
Add AnswerCallbackQuery

sschepens:
Seems it was missing, this allows for Answering Callbacks and thus get the clock icon removed when action finishes.


zoni:
For what it's worth, this looks good to me.

tucnak:
Ace!

lovung:
Do you have any example of using AnswerCallbackQuery in telebot.v1?

PR #56, opened 2016-07-25
Add EditMessageText

sschepens:
First and basic method for editing messages


bot.go
```diff
@@ -152,6 +152,40 @@ func (b *Bot) ForwardMessage(recipient Recipient, message Message) error {
 	return nil
 }
 
+// EditMessage sends a text message to recipient.
```
tucnak: Look, comment-method mismatch!

PR #65, opened 2016-09-24
Improves a bit documentation for newbies like me

JJ:
I'm rather new to Go, so here are the things I needed to do which might help other newbies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,16 @@
 
 Bots are special Telegram accounts designed to handle messages automatically. Users can interact with bots by sending them command messages in private or group chats. These accounts serve as an interface for code running somewhere on your server.
 
-Telebot offers a convenient wrapper to Bots API, so you shouldn't even care about networking at all. Here is an example "helloworld" bot, written with telebot:
+Telebot offers a convenient wrapper to Bots API, so you shouldn't even
+care about networking at all. Install it with
+
+	go get github.com/tucnak/telebot
+
+(after setting up your `GOPATH` properly).
+
+
+
+Here is an example "helloworld" bot, written with telebot:
 ```go
 package main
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,15 +60,15 @@ package main
 import (
     "log"
     "time"
-
+	"os"
```
tucnak: Please keep it consistent with the other imports

README.md
```diff
@@ -46,15 +60,15 @@ package main
...
     "github.com/tucnak/telebot"
 )
 
 var bot *telebot.Bot
 
 func main() {
     var err error
-    bot, err = telebot.NewBot("SECRET TOKEN")
+    bot, err = telebot.NewBot(os.Getenv("BOT_TOKEN"))
```
tucnak: 👍 Totally makes sense, now I wonder why I didn't come up with this earlier.

README.md
```diff
@@ -37,7 +46,12 @@ func main() {
 ## Inline mode
 
 As of January 4, 2016, Telegram added inline mode support for bots.
-Telebot support inline mode in a fancy manner.
+Telebot support inline mode in a fancy manner. Since you are probably
```
tucnak: I can't see how one is coming out of another. It all started with inline modes and there goes the env var thing?

README.md
```diff
@@ -6,7 +6,16 @@
 
 Bots are special Telegram accounts designed to handle messages automatically. Users can interact with bots by sending them command messages in private or group chats. These accounts serve as an interface for code running somewhere on your server.
 
-Telebot offers a convenient wrapper to Bots API, so you shouldn't even care about networking at all. Here is an example "helloworld" bot, written with telebot:
+Telebot offers a convenient wrapper to Bots API, so you shouldn't even
+care about networking at all. Install it with
```
tucnak: I'd say "You may install it with..", but that's a nitpick.

JJ:
Back to work. Thanks for the feedback!

README.md
JJ: That's the actual program I worked with. Maybe I should have moved it up to the first example...

README.md
JJ: It''s the spaces vs. tabs war all over again.. I'll fix that.

JJ:
Thanks! Great job, BTW.

PR #76, opened 2016-11-10
Replace IQR*.MarshalJSON with InlineQueryResultBase

tucnak:
This commit gets rid of multiple redundant copies of MarhsalJSON()
method present for all existing IQR. It does so by introducing a
new struct called InlineQueryResultBase, which takes care of new
Type (for JSON) and existing duplicated ID field.

InlineQueryResults is a new helper wrapper for []IQR slices. It
implements MarshalJSON, which makes sure IQR.ID and IQR.Type are
set properly (by computing a hash and infering IQR type).

/cc @zoni

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -95,8 +95,8 @@ func (results *InlineQueryResults) MarshalJSON() ([]byte, error) {
 			result.SetID(strconv.FormatUint(hash, 16))
 		}
 
-		if err := infereIQR(result); err != nil {
-			return nil, fmt.Errorf("telebot: can't infere type of IQR #%d: %s",
+		if err := inferIQR(result); err != nil {
+			return nil, fmt.Errorf("telebot: can't infer type of IQR #%d: %s",
 				i, err)
 		}
 	}
@@ -104,7 +104,7 @@ func (results *InlineQueryResults) MarshalJSON() ([]byte, error) {
 	return json.Marshal([]InlineQueryResult(*results))
 }
 
-func infereIQR(result InlineQueryResult) error {
+func inferIQR(result InlineQueryResult) error {
 	switch r := result.(type) {
 	case *InlineQueryResultArticle:
 		r.Type = "article"
--- a/inline.go
+++ b/inline.go
@@ -79,6 +79,10 @@ type InlineQueryResult interface {
 // InlineQueryResults is a slice wrapper for convenient marshalling.
 type InlineQueryResults []InlineQueryResult
 
+// MarshalJSON makes sure IQRs have proper IDs and Type variables set.
+//
+// If ID of some result appears empty, it gets set to a new hash.
+// JSON-specific Type gets infered from the actual (specific) IQR type.
 func (results *InlineQueryResults) MarshalJSON() ([]byte, error) {
 	for i, result := range *results {
 		if result.GetID() == "" {
--- a/inline.go
+++ b/inline.go
@@ -1,6 +1,8 @@
 package telebot
 
 import (
+	"encoding/json"
+	"fmt"
 	"hash/fnv"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


inline.go
```diff
@@ -70,18 +72,67 @@ type QueryResponse struct {
...
+	return json.Marshal([]InlineQueryResult(*results))
 }
 
-// hashInlineQueryResult calculates the 64-bit FNV-1 hash of an
-// inline query result.
-func hashInlineQueryResult(result InlineQueryResult) (string, error) {
-	hash, err := hashstructure.Hash(result, inlineQueryHashOptions)
-	if err != nil {
-		return "", err
+func infereIQR(result InlineQueryResult) error {
```
zoni: I think this function is meant to be named "inferIQR" rather than "infereIQR".

tucnak:
I think this PR is a pretty big deal. It makes public API much more concise and logical.

Btw, I just made some commits to master, probably backwards incompatible. I did rename a couple fields... I did my best to keep it neutral though. One of the changes is introduction of type enums, such as `ChatType` and `EntityType` for good.

I'd love if you could test this pull request ASAP. (Nothing urgent though)

zoni:
I took this for a spin today. Good news, it all worked fine once my bot was adjusted to the new API. Here's the diff of what I had to do:

```diff
diff --git a/lycheebot.go b/lycheebot.go
index 7e1b050..7a74248 100644
--- a/lycheebot.go
+++ b/lycheebot.go
@@ -90,29 +90,28 @@ func (lb *LycheeBot) handleQuery(query telebot.Query) {
 			log.Debug("Fetched images from Lychee", "count", len(images))
 		}
 
 		results := make([]telebot.InlineQueryResult, len(images))
 		for i, image := range images {
 			if image.Type == "image/gif" {
 				results[i] = &telebot.InlineQueryResultGif{
-					ID:        strconv.FormatInt(image.ID, 10),
 					GifURL:    fmt.Sprintf("%s/uploads/big/%s", lb.config.Lychee.BaseURL, image.URL),
 					ThumbURL:  fmt.Sprintf("%s/uploads/thumb/%s", lb.config.Lychee.BaseURL, image.ThumbURL),
 					GifWidth:  image.Width,
 					GifHeight: image.Height,
 				}
 			} else {
 				results[i] = &telebot.InlineQueryResultPhoto{
-					ID:          strconv.FormatInt(image.ID, 10),
 					PhotoURL:    fmt.Sprintf("%s/uploads/big/%s", lb.config.Lychee.BaseURL, image.URL),
 					ThumbURL:    fmt.Sprintf("%s/uploads/thumb/%s", lb.config.Lychee.BaseURL, image.ThumbURL),
 					PhotoWidth:  image.Width,
 					PhotoHeight: image.Height,
 				}
 			}
+			results[i].SetID(strconv.FormatInt(image.ID, 10))
 		}
 
 		if len(results) == 0 {
 			nextOffset = ""
 		} else {
 			nextOffset = strconv.Itoa(offset + len(results))
 		}
```

That said, it feels a little weird being able to set all the attributes of an InlineQueryResult *except* for the `ID` field, which must be set later by a call to `SetID()`. Personally I'd prefer if it remained possible to set it when the struct is initialized, which has the benefit that it keeps the original API the same.

But I could totally live with this, it's a minor inconvenience.

tucnak:
I see. Since all IQRs embed IQRBase now, its `ID` field can't be set within struct init statement unfortunately.

Should we push it at the moment thought? I suspect it might cause some broken builds... What do you think?

zoni:
> I see. Since all IQRs embed IQRBase now, its ID field can't be set within struct init statement unfortunately.

Yeah, you're right. I see no way around that. Truthfully, the automatic hashing should be okay in most use-cases so I think it's not too bad to do a separate function call if you want total control over the ID.

> Should we push it at the moment thought? I suspect it might cause some broken builds... What do you think? 

Indeed, I believe it might. Though thankfully the breakage is quite minimal and easily fixed. It is (partly) for this reason that I opened https://github.com/tucnak/telebot/issues/78. It wouldn't avoid breakage in this instance, but would help going forward.

I don't know what other plans you have, if you think there might be more breaking changes in the future than it might make sense to wait and do them all at once, otherwise, perhaps just bite the bullet and deal with any angry users. :smile: (adding a clear changelog and/or section on upgrading to the readme with a note on what to do to fix it could go a long way).

PR #89, opened 2017-08-19
Simple pluggable serve-and-handle API

tucnak:
This is my take on serve-and-handle API, which I think we can introduce effortlessly. It doesn't brake any existing API. It also uses radix tree matching for path serving. Here's the sample of proposed API:

```go
bot, err := telebot.NewBot(os.Getenv("BOT_TOKEN"))
if err != nil {
	log.Fatalln(err)
}

bot.Handle("/start", func (c telebot.Context) {
	b, m := c.Bot, c.Message

	b.SendMessage(m.Sender, "Hello!", nil)
})

messages := make(chan telebot.Message, 100)
bot.Listen(messages, 1*time.Second)

for message := range messages {
	bot.Serve(&message)
}
```

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -21,7 +21,5 @@ func (b *Bot) Serve(msg *Message) {
 
 	if endpoint, ok := value.(Handler); ok {
 		endpoint(Context{b, msg})
-	} else {
-		panic("telebot: couldn't find a route")
 	}
 }
--- a/handlers.go
+++ b/handlers.go
@@ -1,0 +1,27 @@
+package telebot
+
+// Default handler prefix.
+const Default string = ""
+
+type Context struct {
+	Bot *Bot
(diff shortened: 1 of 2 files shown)


irgendwr:
Shouldn't `bot.Serve(&message)` be `bot.Serve(message)` after f7a78ef?

handlers.go
```diff
@@ -0,0 +1,27 @@
+package telebot
+
+// Default handler prefix.
+const Default string = ""
```
irgendwr: This constant is never used

tucnak: I added it for convenience, so I could write:
```go
b.Handle(tb.Default, func(c Context) {})
```

irgendwr: Ah makes sense.

PR #108, opened 2017-12-08
Fixes a nil pointer dereference

martoche:
First time using telebot, I've added a handle to the `OnAddedToGroup` event:

	b.Handle(tb.OnAddedToGroup, func(m *tb.Message) {
		fmt.Printf("%+v\n", m)
	})

and I get a nil pointer dereference error when I create and new group in Telegram with me and the bot. It looks like there is a missing `!= nil` check that this PRs fixes.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -190,9 +190,9 @@ func (b *Bot) incomingUpdate(upd *Update) {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


tucnak:
Cheers!

PR #115, opened 2017-12-30
[BUGFIX] Fixed could not send message to private channel

gSpera:
Sorry for poor English.
I was developing my bot which will send messages to private channels(where it is admin).
I edit _ChatByID_ which now check if the Chat has an username(which is the public link).
If the username is present the channel is public, if not it is private.

I also added the _ChatChannelPrivate_ which is used instead of _ChatChannel_ when channel is private.
I'am not sure if it is 100% correct but it works for me 😄


tucnak:
Cheers!

PR #117, opened 2018-01-04
Added setting for Proxy using.

MarAvFe:
This setting allows to set a Proxy string for the bot to use. 

This way, on creating the bot I can add it like:
```go
b, err := tb.NewBot(tb.Settings{
	Token:  "myt0k3n",
        ...
	Proxy:  "http://proxyhost:port",
})
```

And behind, it will create a client with a transport using the proxy as `client.Post()`. If Proxy were not defined, usual `http.Post()` would be used.

Similar implementation could be used for Bot.sendFiles(). But I haven't had time to try and test an example of those, so it's pending.

### DIFF
--- a/api.go
+++ b/api.go
@@ -30,7 +30,7 @@ func (b *Bot) Raw(method string, payload interface{}) ([]byte, error) {
 
 	var err error
 	resp := &http.Response{}
-	if strings.Compare(b.Proxy, "") != 0 {
+	if b.Proxy == "" {
 		proxyURL, e := url.Parse(b.Proxy)
 		if e != nil {
 			return []byte{}, errors.Wrap(e, "Proxy parsing failed")
@@ -59,7 +59,7 @@ func (b *Bot) sendFiles(
 	method string,
 	files map[string]string,
 	params map[string]string) ([]byte, error) {
-	// ---
+
 	body := &bytes.Buffer{}
 	writer := multipart.NewWriter(body)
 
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -19,17 +20,31 @@ import (
...
 
 	var buf bytes.Buffer
 	if err := json.NewEncoder(&buf).Encode(payload); err != nil {
 		return []byte{}, wrapSystem(err)
 	}
 
-	resp, err := http.Post(url, "application/json", &buf)
+	var err error
+	resp := &http.Response{}
+	if strings.Compare(b.Proxy, "") != 0 {
```
tucnak:
```go
if b.Proxy != "" {
```

bot.go
```diff
@@ -45,6 +46,7 @@ type Bot struct {
 	Token   string
 	Updates chan Update
 	Poller  Poller
+	Proxy   string
```
tucnak, line 49:
```go
// Proxy is a full address to redirect every request into.
```

tucnak (changes requested):

tucnak:
Can you please do it for files too? I wouldn't like to merge inconsistent code, but once it behaves as expected (proxy set => all data must go through the proxy).

bot.go
MarAvFe: I'm not quite sure what to write here to make it better

MarAvFe:
Sure, will do as soon as I can

MarAvFe:
Better late than never. There it is for `sendFiles()` too.

MarAvFe:
Those commits (`04188ff`and`b2dcd74`) are the same. It's just the same merge, that dupped because I didn't wait the first to refresh and repeated it. Dunno why the second fails Travis test. I had some trouble reverting it. Feel free to revert it yourself and then PR will be ready.
Thanks for the project and for letting me help.

turboezh:
Hi.
Does it really needed? Default transport supports HTTP_PROXY env var from the box.
https://golang.org/pkg/net/http/#RoundTripper

irgendwr:
@tucnak This can be closed.

tucnak:
Oh yeah, definitely. Cheers mate. @MarAvFe closing this in favour of #132

PR #119, opened 2018-01-15
simplified unnecessary if

irgendwr:
You can remove this if and just directly return err since it's nil if there is no error.


tucnak:
Cheers!

PR #170, opened 2018-11-02
Add Channel sendable

rdyv:
Usage:
```
var toChannel tg.Recipient = tg.RecipientType{Channel: "@myChannel"}
c.TG.Send(toChannel, message, &tg.ReplyMarkup{
	InlineKeyboard: inlineArr})
```

### DIFF
--- a/sendable.go
+++ b/sendable.go
@@ -12,6 +12,16 @@ type Recipient interface {
 	Recipient() string
 }
 
+// RecipientType - RecipientType
+type RecipientType struct {
+	Channel string `json:"channel"`
+}
+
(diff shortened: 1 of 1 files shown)


stek29:
I'm not sure it should be named "Channel" and not something like "Username".
Also, why is type named `RecipientType` and not something like `RecipientChannel`?
Just `Foo - Foo` isn't enough -- it might pass linters, but it doesn't really help to have such docstring.

And finally, I'm not even sure this should be in library. it's 6 lines of code, and not something most bots need.

stek29:
Maybe it should be added to

PR #175, opened 2018-12-12
Add Custom API URL

idexter:
I think it can be useful in some situations. For example in Russia Telegram DNS is blocked by government. With custom URL we can run HTTP proxy outside Russia and connect though this proxy to API.


tucnak:
Very reasonable, thank you!

PR #176, opened 2018-12-15
Send thumbnails (but not in albums)

stek29:
As it turns out, TG only processes thumbnails if they're uploaded as multipart with field name being `thumb`.

Sending thumbnails in albums is still a mystery.


tucnak:
Soz for a late merge. Good finding

Issue #191, opened 2019-02-11
Escape Telegram Markdown Characters


equt:
Telegram doesn't escape the markdown characters, which might somehow cause the problem. And rather than escape these characters before call the `Send` or `Reply` functions, perhaps it's more elegant to place it inside the telebot and call when the `telebot.ModeMarkdown` is set.

P.S, Only four special characters need to be escaped, ``[ ` * _``.

equt:
Well, this seems to be my fault. Adding this would cause all tokens to be escaped, so this has to be done by the caller.

PR #192, opened 2019-02-11
Replace Telegram markdown characters

equt:
See Issue #191 

> Telegram doesn't escape the markdown characters, which might somehow cause the problem. And rather than escape these characters before call the `Send` or `Reply` functions, perhaps it's more elegant to place it inside the telebot and call when the `telebot.ModeMarkdown` is set.


equt:
See #191

PR #194, opened 2019-02-19
Added support for editMessageMedia.

23rd:


tucnak:
It's pretty good, actually. Sorry for reviewing it rather late!

tucnak:
Thank you!

PR #196, opened 2019-02-26
Fix function comments based on best practices from Effective Go

AngusReid:
Fix function comments following [Effective Go Guidelines](https://golang.org/doc/effective_go.html#commentary)


stek29:
sorry, I've fixed it in a770d6d95118216c7421fc01bd1a2f181564b9e9 along with other issues.
sorry for the long wait and thanks for PR :)

Issue #178, opened 2018-12-17
Setting Content directly to Result


phrozen:
Hi, I'm trying to send some simple **Results** to an inline Query, but I want to be able to set the **ParseMode** to _Markdown_ or _HTML_ in an **ArticleResult**, but I can't set the embedded struct setting, as there is a shortcut Text which says it is mutually exclusive to that. How can I set it?

I tried creating a custom type that satisfies **Result**, but then **inferIQR** is called and totally wrecks my custom type (even if I set the **r.Type** manually to 'article'). Is this intended? I have checked most of the code and can't find a simple way to set my own **InputTextMessageContent** to a known type cause the **Content** field is shadowed.

demget:
Fixed in #225

PR #198, opened 2019-03-24
EditReplyMarkup, switch_inline_query_current_chat, and setters for ResultBase

demget:
### EditReplyMarkup
Wrapper for [editMessageReplyMarkup](https://core.telegram.org/bots/api#editmessagereplymarkup) method.

### InlineButton
I added `switch_inline_query_current_chat` field to InlineButton struct. Note, that you don't need `omitempty` in json tag. If you pass other parameter like `data`, `url` or `switch_inline_query`, Telegram API ignores `switch_inline_query_current_chat`. So it won't cause any conflicts, but will be very useful if you don't want to insert any text in user's input field.

### Setters for ResultBase
It's a bit confusing, when you want to pass custom `Content` to your result (also see #178). I think `ResultBase` don't need pointer to `InputMessageContent`, which is already an interface. Additional setters can solve this problem. I also added `SetReplyMarkup `setter, so now any interaction with `ResultBase` is completely hided.

Example:
```go
// Instead of:
result := &tb.ArticleResult{
	ResultBase: tb.ResultBase{
		Content: new(tb.InputMessageContent),
	},
	// ... other fields
}
*result.Content = &tb.InputTextMessageContent{
	Text: "Some text",
	// ... other fields
}

// I can do this:
result.SetContent(&tb.InputTextMessageContent{
	Text: "Some text",
	// ... other fields
})
```

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -91,14 +91,14 @@ type Settings struct {
 type Update struct {
 	ID int `json:"update_id"`
 
-	Message           *Message  `json:"message,omitempty"`
-	EditedMessage     *Message  `json:"edited_message,omitempty"`
-	ChannelPost       *Message  `json:"channel_post,omitempty"`
-	EditedChannelPost *Message  `json:"edited_channel_post,omitempty"`
-	Callback          *Callback `json:"callback_query,omitempty"`
-	Query             *Query    `json:"inline_query,omitempty"`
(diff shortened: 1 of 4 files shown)


demget:
I've updated the branch. Can you review my PR please?

PR #199, opened 2019-03-25
Add a purge function - Deletes a range of messages

ghost:
Some telegram bots have this, so I implemented this here so people can easily add this functionality to their bot.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -207,19 +207,12 @@ func (b *Bot) Len(chat *Chat) (int, error) {
 // Purge: Deletes a range of messages.
 func (b *Bot) Purge(chat *Chat, start *Message, end *Message) error {
 	// The ID grows by 1 every message so we'll use a for loop and add 1 every run
-	startID := start.ID
-	endID := end.ID
-	for endID>=startID {
-		startIDString := strconv.Itoa(startID) // convert to string because params is a string map
-		params := map[string]string{
-			"chat_id":    strconv.FormatInt(chat.ID, 10),
-			"message_id": startIDString,
-		}
-		_ ,err := b.Raw("deleteMessage", params)
-		if err != nil {
+	msg := start
+	for msg.ID <= end.ID {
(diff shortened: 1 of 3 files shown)


demget:
Hi! The idea of the function is good, but I think that it shouldn't be a part of telebot package. A bot can exceed limits of API or you will need to add intermediate logic for each iteration.

Also your code has several problems:
1) You should use formatting tool for Go code before publishing it.
2) Don't forget about rules of documenting functions.
3) The code can be way clearer, for example:
```go
// Purge deletes a range of messages.
func (b *Bot) Purge(start *Message, end *Message) error {
	msg := *start
	for msg.ID <= end.ID {
                if err := b.Delete(&msg); err != nil {
                        return err
                }
                msg.ID++
	}
	return nil
}
```


If you really need additional functionality, you can implement it on your side like that:
```go
type Bot struct {
        *tb.Bot
}

func (b *Bot) Purge(start *tb.Message, end *tb.Message) error {
        // ...
}
```

ghost:
@demget Thanks for the tips, but how is your code different from:
```// Purge: Deletes a range of messages.
func (b *Bot) Purge(chat *Chat, start *Message, end *Message) error {
	// The ID grows by 1 every message so we'll use a for loop and add 1 every run
	msg := start
	for msg.ID <= end.ID {
		if err := b.Delete(msg); err != nil {
			return err
		}
		msg.ID++
	}
	return nil
}

demget:
@asafniv At the first line you are just storing a pointer to `start` in your variable. When you increment `msg.ID`, it will mutate original `start`. So it's better to get value (so copy it) of pointer like that: `msg := *start` and do things with copy.

tucnak:
I agree with @demget a lot that this is very much of an overkill for telebot itself.

Also, I wouldn't be so sure that IDs can be trusted to progress like that. I'm pretty sure I've noticed cases when users would delete some of their messages and then Telegram would assign their old IDs to these new messages. Unreliable.

PR #204, opened 2019-05-13
- Fix form feed character bug.

mgorunuch:
Bug
image

image
Current regular expression.

image
Changed to this regular expression.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -147,7 +147,7 @@ func (b *Bot) Handle(endpoint interface{}, handler interface{}) {
 
 var (
 	cmdRx   = regexp.MustCompile(`^(\/\w+)(@(\w+))?(\s|$)(.+)?`)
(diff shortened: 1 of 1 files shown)


stek29:
Go regexp is different from JS regex.

```
gore> :import regexp
gore> onef := regexp.MustCompile(`\f([\w.]+)$`)
gore> twof := regexp.MustCompile(`\\f([\w.]+)$`)
gore> onef.FindAllString("\fhi", -1)
[]string{
  "\fhi",
}
gore> twof.FindAllString("\fhi", -1)
[]string{}
```

stek29:
actually, I'm wrong here: JS regex is likely the same.

`\f` is not two symbols (\ and f), it's one symbol (Formfeed Page Break).
When you're putting \f into that regex checker, you have two chars in your text (\ and f), and not one char (\x0c).

notice how data[0] is compared to `'\f'` -- that's also one character and not a string of two characters.

PR #208, opened 2019-05-19
Add is_bot to unmarshal

dhalucario:
I havent tested it yet because its late over here in germany but it kidna seems like a no-brainer honestly.


tucnak:
I can't believe it took me so long to accept this. Sorry for the wait, I must be blind completely.

dhalucario:
Its fine lol. Thanks for accepting.

PR #211, opened 2019-07-11
Inline mode article parse mode

lynkas:
So I tried to send inline message with markdown but failed... and I did not found how... so add a parse to the struct and it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,7 @@ of `QueryResponse`.
 # Contributing
 
 1. Fork it
-2. Clone it: `git clone https://github.com/tucnak/telebot`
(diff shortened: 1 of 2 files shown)


demget:
You should name it `ParseMode` instead of `Parse`. Final code should look like this:
```go
// Optional. Send Markdown or HTML, if you want Telegram apps to show
// bold, italic, fixed-width text or inline URLs in your bot's message.
ParseMode string `json:"parse_mode,omitempty"`
```

demget:
This and other missed fields is added in #229 which, I think, is better.

PR #213, opened 2019-07-23
Fixes last offset problem in getUpdates.

ebarkhordar:
I try it with another messenger with same as telegram API and it takes last update repetitively so bot responses many times to one single update. I can fix it with removing offset from getUpdates params when it is zero. Telegram python SDK dose the same things with offset.

### DIFF
--- a/api.go
+++ b/api.go
@@ -193,6 +193,9 @@ func (b *Bot) getUpdates(offset int, timeout time.Duration) (upd []Update, err e
 		"offset":  strconv.Itoa(offset),
 		"timeout": strconv.Itoa(int(timeout / time.Second)),
 	}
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -193,6 +193,9 @@ func (b *Bot) getUpdates(offset int, timeout time.Duration) (upd []Update, err e
 		"offset":  strconv.Itoa(offset),
 		"timeout": strconv.Itoa(int(timeout / time.Second)),
 	}
+	if offset != 0 {
+		params["offset"] = strconv.Itoa(offset)
```
tucnak, line 197: This code is reduntant, see line no. 193

PR #220, opened 2019-09-17
Avoid using fmt.Sprintf for better performance

mitinarseny:
Hello!

Firstly, let me thank you for developing such a convenient library. I use it for [telego](https://github.com/mitinarseny/telego) and other projects of mine. 

Recently, I've noticed that [telebot](https://github.com/tucnak/telebot) uses [`fmt.Sprintf`](https://golang.org/pkg/fmt/#Sprintf) function, when it is possible to jut concatenate strings using `+`. This would badly affect the speed of the package since `fmt.Sprintf` is [slower](https://dev.to/pmalhaire/concatenate-strings-in-golang-a-quick-benchmark-4ahh) than `+`.

So here is a PR that is intended to slightly increase performance of this excellent package.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -545,7 +545,7 @@ func (b *Bot) SendAlbum(to Recipient, a Album, options ...interface{}) ([]Messag
 		} else if f.FileURL != "" {
 			mediaRepr = f.FileURL
 		} else if f.OnDisk() || f.FileReader != nil {
-			mediaRepr = fmt.Sprintf("attach://%d", i)
+			mediaRepr = "attach://" + strconv.Itoa(i)
 			files[strconv.Itoa(i)] = *f
 		} else {
 			return nil, errors.Errorf(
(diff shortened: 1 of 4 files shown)


bot.go
```diff
@@ -1062,7 +1061,7 @@ func (b *Bot) StopLiveLocation(message Editable, options ...interface{}) (*Messa
 	messageID, chatID := message.MessageSig()
 
 	params := map[string]string{
-		"chat_id":    fmt.Sprintf("%d", chatID),
```
stek29, line 1065: Why `FormatInt` and not `Itoa`?

mitinarseny: @stek29 
Because `Message.MessageSig()` returns `int64`, but `strconv.Itoa` accepts only `int`. Anyway, `strconv.Itoa` [uses](https://golang.org/src/strconv/itoa.go?s=1028:1051#L24) `strconv.FormatInt`:
```go
// Itoa is equivalent to FormatInt(int64(i), 10).
func Itoa(i int) string {
	return FormatInt(int64(i), 10)
}
```

tucnak:
Perfect! And thank you for the kind words about the library. It's still a long way to go, but I'm proud to hear things like these from people like you, and you personally.

Cheers,
Ian

Issue #203, opened 2019-04-26
Sendable objects issues


demget:
I found some bugs while working with files in telebot.

### Object's caption
```go
doc := &tb.Document{
	File:    tb.FromURL("https://via.placeholder.com/300x150"),
	Caption: "caption",
}

bot.Send(&tb.User{ID: userID}, doc) // with "caption"
bot.Send(&tb.User{ID: userID}, doc) // no caption
```

### FileName
```go
bot.Send(&tb.User{ID: userID}, &tb.Document{
	File:     tb.FromReader(reader),
	FileName: "file.txt", // sends "empty"
})
```

demget:
Fixed by #226

PR #226, opened 2019-09-30
Media objects improvements

demget:
Related to #203.

### File name
Now you can pass your custom file name to an audio, video or document. It's very helpful feature when you work with files. A commit changes `addFileToWriter`, so I also refactor the code a bit to make it good-looking.

### Caption
Another bug was related to captions. When you send any type of media, its caption disappears in your original object. It happens after `stealRef` section, where is value of receiver sets to returned message's media. But actual caption is storing in the message, not media. _It's better to check caption-related changes one more time, since I hadn't enough time to test it perfectly._

### DIFF
--- a/api.go
+++ b/api.go
@@ -8,7 +8,6 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


tucnak:
Thank you!

PR #229, opened 2019-10-28
Update InlineQueryResult types

demget:
There were a lot of changes in Bot API since the first implementation of an inline API, so I updated all the InlineQueryResult types.

Another issue was with ParseMode in `InputTextMessageContent` struct. You always had to convert original ParseMode type to string, as defined in the `InputTextMessageContent`. Changing type of this field from `string` to `tb.ParseMode` breaks backwards compatibility. So, the best and safe solution in this case is making `ParseMode` type identical, not different: `type ParseMode = string`.

```go
result.SetContent(&tb.InputTextMessageContent{
    // instead of this:
    ParseMode: string(tb.ModeMarkdown),
    // now you can do this:
    ParseMode: tb.ModeMarkdown,
})
```

### DIFF
--- a/inline_types.go
+++ b/inline_types.go
@@ -65,6 +65,12 @@ type ArticleResult struct {
 
 	// Optional. URL of the thumbnail for the result.
 	ThumbURL string `json:"thumb_url,omitempty"`
+
+	// Optional. Width of the thumbnail for the result.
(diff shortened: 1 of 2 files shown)


stek29:
I've realized it wasn't in develop yet, and it doesn't break anything, should've been merged long time ago.

PR #230, opened 2019-11-03
Small fixes and improvements

demget:


message.go
```diff
@@ -180,6 +180,9 @@ type Message struct {
 	// in this field will not contain further ReplyTo fields even
 	// if it is itself a reply.
 	PinnedMessage *Message `json:"pinned_message"`
+
+	// Inline keyboard attached to the message.
+	ReplyMarkup InlineKeyboardMarkup `json:"reply_markup"`
```
stek29, line 185: It's used as pointer in other places in this library
But since it's just a struct with one slice inside using it as value is better imo

stek29:
Thank you.

I've merged it manually to have EditCaption and Message.ReplyMarkup in separate commits, I hope it's ok.

PR #237, opened 2019-12-08
Callback's inline message

demget:
This branch allows to call `Edit`, `EditCaption`, ... methods with inline messages. I added helper field `InlineID` to `Message` struct, which indicates that the message is sent by inline mode. Extended `MessageSig` will return this id with zero `chat_id`.

Inline message was handled in edit functions like so:
```go
// If inline message.
if chatID == 0 {
	params["inline_message_id"] = messageID
} else {
	params["chat_id"] = strconv.FormatInt(chatID, 10)
	params["message_id"] = messageID
}
```

..but there wasn't any built-in interface that represents inline message with valid `MessageSig` implementation.

And now, you can do something like this with no problems:
```go
func OnCallback(c *tb.Callback) {
    b.Edit(c, "new text") // even if it's inline message
}
```

I also added helper function `IsInline` for `Callback` type.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -312,8 +312,13 @@ func (b *Bot) incomingUpdate(upd *Update) {
 
 	if upd.Callback != nil {
 		if upd.Callback.Data != "" {
-			data := upd.Callback.Data
+			if upd.Callback.MessageID != "" {
+				upd.Callback.Message = &Message{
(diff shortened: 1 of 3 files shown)


tucnak:
Thank you! Sorry for the long wait.

PR #250, opened 2020-02-29
Add sendAnimation

fullpipe:
Added:
- https://core.telegram.org/bots/api#animation
- https://core.telegram.org/bots/api#sendanimation


tucnak:
Sorry for the long wait! Code looks great actually, tested it—works fine.Thank you!

PR #251, opened 2020-02-29
Add support parse_mode for albums photo

dzen-it:
https://core.telegram.org/bots/api#inputmediaphoto
The "parse_mode" option was not processed in the albums photos.
Added "omitempty" for "caption" because it's optional for request.
Fixed.


stek29:
Thank you

PR #256, opened 2020-03-28
Spellcheck/reword

kn100:
I wanted to read the entire readme for this as I am planning on using it - so I thought I'd spellcheck/reword at the same time! Thanks for the great lib.


tucnak:
Cheers!

PR #257, opened 2020-03-30
Implement poll API

setval:


demget:
Thank you!

poll.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (p *Poll) IsRegular() bool {
+	return p.Type == "regular"
+}
+
+func (p *Poll) IsQuiz() bool {
+	return p.Type == "quiz"
+}
+
+func (p *Poll) AddAnswers(text ...string) {
```
stek29, line 43: why is it named AddAnswers, if it actually adds Options?

setval: Initially, it was planned that options (PollOption) only contains Text, so I decided to call the method this way (that is, the VoterCount field is not filled in). But now I really understand that perhaps another name would be acceptable. I reserve the right for collaborators.

stek29: @demget maybe it’s a good idea to change it before develop is merged?

PR #269, opened 2020-04-25
Dice API support

0xfedcafe:


demget (changes requested):
Please do not forget about gaps between logical blocks of code. 
Send method should look like this:

```go
}

// Send delivers dice through bot b to recipient.
func (d *Dice) Send(b *Bot, to Recipient, opt *SendOptions) (*Message, error) {
	params := map[string]string{
		"chat_id": to.Recipient(),
		"emoji":   string(d.Type),
	}
	embedSendOptions(params, opt)

	data, err := b.Raw("sendDice", params)
	if err != nil {
		return nil, err
	}

	return extractMessage(data)
}
```

demget:
Thank you!

demget (commented):
Ok

PR #270, opened 2020-04-26
Added balancer to avoid Telegram API limits

Pask00:
A simple implementation to avoid API limits.
What do you think?

### DIFF
--- a/api.go
+++ b/api.go
@@ -18,6 +18,13 @@ import (
 
 // Raw lets you call any method of Bot API manually.
 func (b *Bot) Raw(method string, payload interface{}) ([]byte, error) {
+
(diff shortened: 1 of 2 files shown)


demget:
I think this implementation is not the best and quite awkward. Firstly, for such thing as load balancer you should add basic tests at least. We already have some thoughts about good balancer implementation in future releases and it's more complicated than simple requests counter. Anyway thanks for your try, but as for me, your changes isn't satisfactory enough to merge.

demget:
Also, your manipulations with `callsCounter` are not thread safe.

stek29:
Hard coding limits is also a bad idea

PR #275, opened 2020-04-27
ref: set omitemty for InlineQueryChat to avoid telegram api limits

naderAbolfazli:


demget:
It breaks backwards compatibility, because you won't be able to answer inline query with empty  `switch_inline_query_current_chat` param

demget:
Can you elaborate please, what limits are you talking about?

naderAbolfazli:
> Can you elaborate please, what limits are you talking about?
I get it. my problem was using another bot server instead of telegram that not handles this param.

stek29:
@naderAbolfazli are there other servers for BotAPI?

naderAbolfazli:
> @naderAbolfazli are there other servers for BotAPI?

yes. in other messengers

PR #276, opened 2020-04-27
errors: add ErrNotStartedByUser

orivej:
This error happens when the bot attempts to message a user that has not started it. (In my code I need to treat this error like ErrBlockedByUser, by sending a group message that explains why the bot did not send a private message and how to fix this.)

For the reference, see https://github.com/YouTwitFace/telegram-errors/commit/c2332a0495183cad5ccc877c6ab3e92ab36d85ec


demget:
Thank you!

PR #279, opened 2020-04-28
bot: expose ProcessUpdate and add Synchronous mode to support AWS Lambda

orivej:
(Please review individual commits. The last one that adds offline testing mode is incidental.)

AWS Lambda is well suited for running a Telegram bot. (See [Go examples](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html).) To use it I need to make a function that handles one Telegram update and returns. It has to handle the update until completion since Lambda can kill the program as soon as the function passed to lambda.Start returns: this is why we need the synchronous mode that waits for the user handler. This PR makes telebot suitable for such a use:

```go
package main

import (
	"github.com/aws/aws-lambda-go/lambda"
	tb "gopkg.in/tucnak/telebot.v2"
)

func main() {
	bot, _ := tb.NewBot(tb.Settings{Synchronous: true})
	lambda.Start(func(u tb.Update) error {
		bot.ProcessUpdate(&u)
		return nil
	})
}
```


util.go
```diff
@@ -28,6 +28,18 @@ func (b *Bot) deferDebug() {
 	}
 }
 
+func (b *Bot) runHandler(handler func()) {
+	f := func() {
+		defer b.deferDebug()
```
orivej, line 33: Note that without this PR `deferDebug` is deferred only if b.reporter is nil, but it seems that it should be always deferred, since without it no one handles panics in handlers, whereas Settings.Reporter says that it should be called on panic.

demget:
I really liked you added the offline mode! Seems pretty useful.

But do we actually need additional closures like here?
```go
func(b *Bot, handler func(*Message), m *Message) {
     b.runHandler(func() { handler(m) })
}(b, handler, m)
```

orivej:
I've remove the closures, and changed the signature of ProcessUpdate to take Update instead of \*Update since in asynchronous mode it would not be safe to capture the pointer if the user calls ProcessUpdate multiple times with a pointer to the same Update struct (with different content). (I could have kept `func (b *Bot) incomingUpdate(upd *Update)` and added `func (b *Bot) ProcessUpdate(upd Update) { b.incomingUpdate(&upd) }`, but this does not seem useful enough.)

orivej:
Could you point out the changes that do not feel quite right, so that I would think about them some more?

util.go
tucnak: Good catch!

tucnak (changes requested):
The code is solid, the only missing piece is documentation for README.md. This patchi introduces quite a few changes, so we would need to see corresponding docs...

I'll merge this once there is an example clause in README, which outlines how this is supposed to be used in AWS Lambda context. That said, how does Lambda work for you? We've tried it numerous times and quite a few times it did us proud, but oh–my–bills, we figured it must be not worth it for us.

Thank you!

orivej:
I have added a fully functional minimal example (extracted from my bot): this seems more useful than just an outline. For example, the value that the function started with `lambda.Start` receives depends on the kind of AWS API Gateway (HTTP, REST, v1 or v2), so it helps to have a declarative and reproducible AWS configuration alongside the code.

> That said, how does Lambda work for you?

This is my first Lambda project, perhaps I shall tell how it works out in a few months. Lambda seems perfect for my use case because I do not maintain a dedicated server where I could co-host my bot, and because I expect very low usage. It may cost under a cent per month.

orivej:
I have added Synchronous=true to TestBotIncomingUpdate: now all settings are tested and the asserts in its handlers run within the test.

tucnak (approved):

tucnak:
It is solid. Thank you so much and sorry for the wait!

tcurdt:
I want to give this a try - but I don't quite understand how the lambda is being called.
It's not like I we are telling telegram the URL - and there is no poller.

README.md
```diff
@@ -109,7 +109,9 @@ backwards-compatibity.
 
 ## Poller
 Telebot doesn't really care how you provide it with incoming updates, as long
-as you set it up with a Poller:
+as you set it up with a Poller, or call ProcessUpdate for each update (see
```
suntong, line 112: The provided example is somehow removed later on. What happened? Is it no longer working? thx!

suntong: > `tb "gopkg.in/tucnak/telebot.v2"` 

Oh, it's for V2 only, and not working in V3 any more?

PR #283, opened 2020-05-05
fix errorRx to match other APIError

guest6379:
The pattern `{.+"error_code":(\d+),"description":"(.+)"}` can match the result of json:
```json
{"ok":false,"error_code":400,"description":"Bad Request: message to edit not found"}
```

but cannot match the one below:
```json
{"ok":false,"error_code":429,"description":"Too Many Requests: retry after 10","parameters":{"retry_after":10}}
```
So change it to `{.+"error_code":(\d+),"description":"(.+).*"}`, which can prevent the ` nil pointer error ` from happening

https://github.com/tucnak/telebot/blob/319196b8471500baece223be6acfc94a04280226/util.go#L48-L56

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -72,3 +72,17 @@ func TestRaw(t *testing.T) {
 	_, err = b.Raw("testUnknownError", nil)
 	assert.EqualError(t, err, "telegram unknown: unknown error (400)")
 }
+
+func TestExtractOK(t *testing.T) {
+	data := []byte(`{"ok":true}`)
+	assert.Nil(t, extractOk(data))
+
+	data = []byte(`{"ok":true,"result":{}`)
+	assert.Nil(t, extractOk(data))
+
+	data = []byte(`{"ok":false,"error_code":400,"description":"Bad Request: message to edit not found"}`)
+	assert.EqualError(t, extractOk(data), extractOk(data).Error())
+
+	data = []byte(`{"ok":false,"error_code":429,"description":"Too Many Requests: retry after 10","parameters":{"retry_after":10}}`)
+	assert.EqualError(t, extractOk(data), extractOk(data).Error())
+}
--- a/errors.go
+++ b/errors.go
@@ -45,7 +45,7 @@ func NewAPIError(code int, msgs ...string) *APIError {
 	return err
 }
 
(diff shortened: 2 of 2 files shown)


demget:
Good catch, but actually your expresssion is also wrong:
`{.+"error_code":(\d+),"description":"(.+).*"}`

I guess `.*` pattern should be after description quotes:
`{.+"error_code":(\d+),"description":"(.+)".*}`

demget:
Given this situation, I think we should have some tests for `extractOk` to be sure that regexp is working well. It'll be awesome if you can write some. Something like `TestExtract` in `api_test.go` file.

Also, I fixed middleware poller test to satisfy travis build. Now it should be fine.

guest6379:
> Good catch, but actually your expresssion is also wrong:
> `{.+"error_code":(\d+),"description":"(.+).*"}`
> 
> I guess `.*` pattern should be after description quotes:
> `{.+"error_code":(\d+),"description":"(.+)".*}`

Sorry for my mistake before, thank you for help, in this commit, I try to add some tests for extractOK.

demget:
Tests are ok. But maybe it's clearer to use `assert.EqualError` instead of `assert.EqualValues`?

guest6379:
> Tests are ok. But maybe it's clearer to use `assert.EqualError` instead of `assert.EqualValues`?

Thank you for pointing this out, I'm trying to meet this.

api_test.go
```diff
@@ -72,3 +72,17 @@ func TestRaw(t *testing.T) {
...
+
+func TestExtractOK(t *testing.T) {
+	data := []byte(`{"ok":true}`)
+	assert.Nil(t, extractOk(data))
+
+	data = []byte(`{"ok":true,"result":{}`)
+	assert.Nil(t, extractOk(data))
+
+	data = []byte(`{"ok":false,"error_code":400,"description":"Bad Request: message to edit not found"}`)
+	assert.EqualError(t, extractOk(data), extractOk(data).Error())
```
stek29, line 84: what's the point of comparing `err` to `err.Error()`? assert.NotNil should be enough. If you want to use EqualError to check what error is actually returned, compare to error from errors.go.

demget:
Similar PR is already merged: #294

PR #293, opened 2020-05-13
payments: update API

setval:
Update Payments API


setval:
Before mergem I want to clarify one detail. Maybe I have written incorrectly by adding a field in Photo with the JSON tag photo_size
https://github.com/tucnak/telebot/blob/76c2cedf6f513e73436fb23638e6f4b80942b2db/media.go#L25-L27

I leave the decision to the owner and collaborators

demget:
Thank you!

PR #294, opened 2020-05-18
errors: fix regular expression

aofei:
Telegram Bot API now returns an extra field for error responses, such as:

```json
{"ok":false,"error_code":429,"description":"Too Many Requests: retry after 8","parameters":{"retry_after":8}}
```

This means that existing regular expression cannot extract errors normally, so it will cause panic.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -45,7 +45,7 @@ func NewAPIError(code int, msgs ...string) *APIError {
 	return err
 }
 
-var errorRx = regexp.MustCompile(`{.+"error_code":(\d+),"description":"(.+)"}`)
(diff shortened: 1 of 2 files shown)


demget:
We have similar #283, but no approved tests. Can you provide some so I will be able to merge your pull request?

aofei:
Sorry, my bad. I forgot to check the PR list, these two PRs are basically the same. 😅

Anyway, I added tests. And I think this is a problem that needs to be fixed ASAP, it caused too many panics.

demget:
According to the contribution guide, you should change a base branch to `develop`

aofei:
> According to the contribution guide, you should change a base branch to `develop`

Done.

demget:
Thank you!

PR #311, opened 2020-06-12
api: fix a huge memory allocation caused by sendFiles

aofei:
I encountered an unexpected large memory leak in a program that uploads large files concurrently. After several days of investigation, I found that it was caused by the `Bot.sendFiles`. At present, the behavior of buffering all content in runtime memory before sending it to the server side is very crazy. It means the memory allocation grows linearly with the file size. So I fixed it by using the `io.Pipe`.


tucnak:
This is very crazy indeed. Good catch, thank you!

PR #313, opened 2020-06-15
api: make sure io.Pipe in sendFiles will be closed

aofei:
This is a supplementary PR for #311.

There is a rare situation that may cause the `io.Pipe` in the `Bot.sendFiles` to not be closed. That is when the `b.client.Post` returns an error before reading any data from the `pipeReader`. Sorry, this is my fault.

And, this PR also fixes duplicate `wrapError`.


tucnak:
Nice catch!

PR #316, opened 2020-06-24
Change param type of embedRights 

### LINKED ISSUE #315: Bad serialization for Rights
Some bot methods have [`Rights`](https://github.com/tucnak/telebot/blob/b9ba8c453575510d2ad20c4335b4441f36f9f6e4/admin.go#L103-L123) included in their params and using `embedRights` to patch `Rights` to request params. However, in `embedRights`, while `Rights` is a struct of `bool`s, `Rights` is patched to a `map[string]string`, which caused values in the map are set to `""`. Whatever the content in `Rights` is, they become `false` in Telegram's view.

Related code snippets:
https://github.com/tucnak/telebot/blob/b9ba8c453575510d2ad20c4335b4441f36f9f6e4/util.go#L188-L191

See also: <https://play.golang.org/p/lJCE0Zu2v36>

Telegram bot API doc section: <https://core.telegram.org/bots/api#chatmember>

It seems the following bot methods are affected: (in my case, using `Restrict` to unset the restriction of a user to send messages is broken)
```go
/* admin.go */
func (b *Bot) Restrict(chat *Chat, member *ChatMember) error
func (b *Bot) Promote(chat *Chat, member *ChatMember) error
/* util.go */
func (b *Bot) SetGroupPermissions(chat *Chat, perms Rights) error
```

I can open a PR on this. What do you think about changing the type of `p` in `embedRights` from `map[string]string` to `map[string]interface{}`?

ghost:
This will close #315.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -111,7 +111,7 @@ func (b *Bot) Unban(chat *Chat, user *User) error {
 func (b *Bot) Restrict(chat *Chat, member *ChatMember) error {
 	prv, until := member.Rights, member.RestrictedUntil
 
-	params := map[string]string{
+	params := map[string]interface{}{
 		"chat_id":    chat.Recipient(),
 		"user_id":    member.User.Recipient(),
 		"until_date": strconv.FormatInt(until, 10),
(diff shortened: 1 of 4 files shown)


demget:
Thank you for your help!

PR #324, opened 2020-10-01
Fix inlinebutton With func

profawk:
the InlineButton With func returned a new button with it's url property mangled (t.Text instead of t.URL)
this caused telegram to respoind with 400


demget:
Good catch, thank you!

PR #338, opened 2020-12-10
Add bowling dice

PaulLesur:


demget:
Oh missed it. What about simply `Bowl` to make it 4-letters as other?

PaulLesur:
Yep I could set it to `Bowl` but what if Telegram add a game with the bowl emoji :bowl_with_spoon: ? :stuck_out_tongue:

demget:
Thanks

PR #340, opened 2020-12-13
bot: fix FileURLByID for local Bot API server

aofei:
According to [this](https://core.telegram.org/bots/api#using-a-local-bot-api-server):

> * Receive the absolute local path as a value of the *file_path* field without the need to download the file after a [getFile](https://core.telegram.org/bots/api#getfile) request.

We can assume that the file whose `FilePath` is an absolute path is from the local disk. So, we shouldn't return `http://` URL for it, `file:///` URL is more appropriate.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


demget:
Thank you, sorry for delay.

PR #343, opened 2020-12-28
Update editable.go

sashker:
Might be an error for the`messageID` type, since it directly contradicts documentation (i.e. we are waiting for int instead of string)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: go
 
 go:
-    - 1.12.x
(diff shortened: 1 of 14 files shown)


demget:
Backward incompatible for `v2`. Maybe it really makes sense for`v3`, I'll think about it.

demget:
Well, it's better for me to make the change in `v3` after the #412 decision so without any conflicts...

PR #360, opened 2021-02-03
Fix error handling in Nil response case in the send Object

mmdaz:
In some cases (I think in the internal error case) `resp.Result` is Nil and the application crashes in the next functions such as Send functions in the Sendable interface.


demget:
I would like to see these cases with nil response then, to ensure it really happens and investigate why.

demget:
As `v2` is no longer supported, I have to close the PR. Consider migrating to `v3`.

Issue #365, opened 2021-02-09
Incorrect handling of UsersJoined


dexter3k:
This piece of code
https://github.com/tucnak/telebot/blob/7a9c471cf32c81870f1bad0fdfc8fa6a646abb89/bot.go#L263
Incorrectly handles UserJoined logic.

When multiple users are added to the chat, UsersJoined will contain list of users joined, but also UserJoined will also contain one of the users from the list. Currently because
```golang
if m.UserJoined != nil {
	b.handle(OnUserJoined, m)
	return
}

if m.UsersJoined != nil {
	for _, user := range m.UsersJoined {
		m.UserJoined = &user
		b.handle(OnUserJoined, m)
	}
	return
}
```
is there, only the first user will be handled.
One should probably use something like this:
```golang
if m.UsersJoined != nil {
	for _, user := range m.UsersJoined {
		m.UserJoined = &user
		b.handle(OnUserJoined, m)
	}
	return
} else if m.UserJoined != nil {
	b.handle(OnUserJoined, m)
	return
}
```

dexter3k:
Now, there seems to be another issue closely related to this, not sure if this better be another issue..

There is a data race in your code here
```golang
for _, user := range m.UsersJoined {
  m.UserJoined = &user
  b.handle(OnUserJoined, m)
}
```

Because b.handle runs handlers asynchronously in some cases, the data at &user will get updated to the last user in the list. Thus all, or some calls to handlers will get same user.

You can try live example of this problem simplified here: https://play.golang.org/p/DbOZgUOzyV1

tucnak:
Fixed in #366

PR #366, opened 2021-02-09
Fixing handling of multiple users joining

dexter3k:
Here is a fix for #365

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -260,16 +260,14 @@ func (b *Bot) ProcessUpdate(upd Update) {
 			return
 		}
 
-		if m.UserJoined != nil {
-			b.handle(OnUserJoined, m)
-			return
-		}
-
 		if m.UsersJoined != nil {
-			for _, user := range m.UsersJoined {
-				m.UserJoined = &user
+			for index := range m.UsersJoined {
+				m.UserJoined = &m.UsersJoined[index]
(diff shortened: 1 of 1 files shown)


bot.go
```diff
@@ -260,17 +260,15 @@ func (b *Bot) ProcessUpdate(upd Update) {
...
-
 		if m.UsersJoined != nil {
-			for _, user := range m.UsersJoined {
-				m.UserJoined = &user
+			for index := range m.UsersJoined {
+				m.UserJoined = &m.UsersJoined[index]
 				b.handle(OnUserJoined, m)
 			}
 			return
+		} else if m.UserJoined != nil {
```
kifirkin: You could just put it like that, as you have a return statement before
```suggestion
		}

		if m.UserJoined != nil {
```

dexter3k:
Looks like this is not all. Somehow my fix worked for me, but not on another machine :/
Same problem stays:
```golang
for index := range m.UsersJoined {
    m.UserJoined = &m.UsersJoined[index]
    b.handle(OnUserJoined, m)
}
```
Here we're not racing on the loop variable anymore, but on UserJoined struct field.
Instead, one has to make a new instance of Message. I think shallow copy is enough.

tucnak:
Great work, thank you!

PR #371, opened 2021-02-22
Expose offline settings for Bot

metalmatze:
I want to write integration tests with the bot myself. I don't plan on re-implementing the command routing etc. 
It would be super helpful to use the offline mode for that externally to the project.
These changes allow my tests to pass successfully.


metalmatze:
Wow. That was fast. :smile: 
Just came here while adding the Go Modules `replace`. Guess I don't need that. :rocket:

PR #375, opened 2021-03-12
Support get/set

### LINKED ISSUE #374: Support Get and Set value in Context
Context interface should have `Get` and `Set` functions for passing values via context.
An example of an interface - https://github.com/labstack/echo/blob/45870c75c3b3fb20edec25cb09dc14ab17df59b9/context.go#L100-L104 

For what? For example, you can create middleware that adds a user to the context:
```go
func User(manager account.Interface) telebot.MiddlewareFunc {
	return func(next telebot.HandlerFunc) telebot.HandlerFunc {
		return func(c telebot.Context) error {
			user, err := manager.GetTelegramUser(context.Background(), int64(c.Sender().ID))
			if err != nil {
				return err
			}
			c.Set("user", user)

			return next(c)
		}
	}
}
```

lcd1232:
Closes #374

### DIFF
--- a/context.go
+++ b/context.go
@@ -2,6 +2,7 @@ package telebot
 
 import (
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


lcd1232:
@tucnak ping

0xfedcafe:
> @tucnak ping

pong

PR #377, opened 2021-03-20
implements voice chat service messages

omar-polo:
Hello,

I wanted to play with the VoiceChat service messages, but telebot lacked the support, hence this PR.
The VoiceChat service messages are part of the bot API v5.1.

I'm not sure of:

 - is it OK to add `voicechat.go` or it's better to stuff those declarations in some other file?
 - The Duration field of the VoiceChatEnded should be a time.Duration instead of a int? (it's the number of seconds)

Cheers!

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -8,20 +8,21 @@ import (
 
 // Rights is a list of privileges available to chat members.
 type Rights struct {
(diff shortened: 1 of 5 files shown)


tucnak:
Thank you!

PR #382, opened 2021-03-30
implement my_chat_member and chat_member

omar-polo:
This adds the support for the new `my_chat_member` and `chat_member` update types present in bot API v5.1.  `my_chat_member` in particular is handy when you want to know if a bot was added (as an admin) to a channel.

Honestly, I have tested only the `my_chat_member` and not the `chat_member`, but I've added that too since it uses the same struct.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -6,6 +6,61 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


demget:
Great job, thank you

Issue #387, opened 2021-04-05
Add support for "caption" in sendVoice


rusq:
Currently, in Voice structure apart from embedded File, there's only [MIME](https://github.com/tucnak/telebot/blob/v2/media.go#L164) member variable (speaking of V2).

Telegram [sendVoice](https://core.telegram.org/bots/api#sendvoice) Bot API now supports sending a string caption (max 1024 bytes) with voice messages.

I'll submit PR shortly.

rusq:
Sample message with caption:
![image](https://user-images.githubusercontent.com/16064414/113559230-1e51dc80-9655-11eb-8185-670101237545.png)

rusq:
Hey, @demget - should submit this to the V3 branch as well, if needed? (haven't checked that branch yet).

PR #388, opened 2021-04-05
Add Caption parameter to Voice structure

rusq:
As per #387 adding a Caption to Voice, to reflect the current [Bot API state](https://core.telegram.org/bots/api#sendvoice)


demget:
Use develop branch as target next time please

demget:
Thank you!

Issue #90, opened 2017-08-20
MessageEntity with EntityMention type broken?


Riketta:
Can't receive User info from MessageEntity instance via such code:
`			for _, v := range message.Entities {
				if v.Type == telebot.EntityMention || v.Type == telebot.EntityTMention { 
					fmt.Println("[Entity]", v)
				}
			}`
Possible output:
> [Entity] {mention 0 8  {0   }}
> [MSG]: From ... in chat #b2: @riketta
> [Entity] {mention 4 8  {0   }}
> [MSG]: From ... in chat #b2: qwe @riketta  aasas

Seems like entity.User contains nothing useful. Sorry if I'm doing smth wrong.

tucnak:
According to Telegram Bot API, `MessageEntity.User` is for users without username only (see https://core.telegram.org/bots/api#messageentity).

For users with username, you can just go and slice it like that: `msg.Text[offset:offset+length]`. It gets trickier for any other mentions though! Since UTF-16 code points may not be 1-byte long, you can't just slice them straight away (strings in Go are immutable []byte and UTF-16 is a variable-length encoding)... I guess we could to introduce a helper function for this very reason.

PR #391, opened 2021-04-11
add Message.EntityText to ease managing MessageEntities

omar-polo:
Hello,

This PR adds an helper function to retrieve the message substring pointed by a MessageEntity.  A MessageEntity contains a Offset and Length fields, but these are "UTF-16 code units" (I'm assuming "code units" means "in code points"), so one cannot safely slice message.Text.  (UTF-16 is a variable-length, multibyte encoding and Go strings are byte slices).

It's easy to verify that, for a message that contains non-ASCII codepoints, the slice `m.Text[e.Offset : e.Offset+e.Length]` picks up the wrong substring.

In #90 the idea of such method was ventilated

Regarding the implementation: it may be possible to do the same thing without converting the whole message.Text to utf-16, but since telegram messages have a (relatively short) maximum length, it's probably worth the simplicity of the implementation.

Returning the empty string if the indices are wrong instead of returning an explicit error is to keep the function (and its utilization) simple.  One should provide a MessageEntity from the very same message, and we kinda can trust telegram to provide correct indices.

Thoughts?


omar-polo:
Sorry, PR sent too fast. I've noticed only now that there is also a CaptionEntities field, so this needs to be adjusted.

What's the better approach?  Provide an helper function not tied to the Message type, providing a similar CaptionEntityText or using m.Caption if m.Text is empty?

omar-polo:
I added a CaptionTextEntity.  I would have preferred something like `entity.Text(m)`, but we have no way to know if an entity refers to the Text or Caption unfortunately :/

omar-polo:
[sorry for spamming, but]

I'm still thinking on what could be a better interface, and I thought of this:

```
func (e *MessageEntity) Text(text string) string
```

it's slightly better than the current proposal because it introduces only one function, it's the caller responsability to correctly pass `Message.Text` or `Message.Caption`.

demget:
A great helper and it's already in `v3`. As `v2` is no longer supported, I have to close the PR. Consider migrating to the `v3`.

omar-polo:
Yep, I forgot about this pr!

thanks :)

PR #392, opened 2021-04-11
use only replay message id instead of whole message in SendOptions

giorgijpopov:
it is not necessary to store whole `ReplyTo` message in `SendOptions`


demget:
The thing is, we can't break backward compatibility for `v2`, especially here in `SendOptions`, since it was from the very beginning.

PR #401, opened 2021-05-14
Add copyMessage

comerc:
https://core.telegram.org/bots/api#copymessage


demget:
Consider resetting the last commit as it's already implemented in #407, so I can merge it.

PR #403, opened 2021-05-18
Add revoke_messages parameter to kickChatMember API method

### LINKED ISSUE #402: API update: revoke_messages parameter in kickChatMember method
See https://core.telegram.org/bots/api#kickchatmember:

> revoke_messages: Pass True to delete all messages from the chat for the user that is being removed. If False, the user will be able to see messages in the group that were sent before the user was removed. Always True for supergroups and channels.

AlexanderMatveev:
Fixes #402

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -136,11 +136,12 @@ func Forever() int64 {
 }
 
 // Ban will ban user from chat until `member.RestrictedUntil`.
-func (b *Bot) Ban(chat *Chat, member *ChatMember) error {
+func (b *Bot) Ban(chat *Chat, member *ChatMember, revokeMessages bool) error {
 	params := map[string]string{
-		"chat_id":    chat.Recipient(),
-		"user_id":    member.User.Recipient(),
-		"until_date": strconv.FormatInt(member.RestrictedUntil, 10),
+		"chat_id":         chat.Recipient(),
+		"user_id":         member.User.Recipient(),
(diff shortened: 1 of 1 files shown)


admin.go
```diff
@@ -136,11 +136,12 @@ func Forever() int64 {
 }
 
 // Ban will ban user from chat until `member.RestrictedUntil`.
-func (b *Bot) Ban(chat *Chat, member *ChatMember) error {
+func (b *Bot) Ban(chat *Chat, member *ChatMember, revokeMessages bool) error {
```
demget, line 139: We can't just add another argument to the widely used `Ban` function, because it will break backward compatibility. And since `revoke_messages` doesn't fit in `ChatMember` structure, the only option is to make `revokeMessages` variadic and if the additional boolean argument was given, to check the length and reassign the local `revoke` variable, before passing it to the map.

demget (changes requested):

demget:
Implemented in #407

admin.go
AlexanderMatveev: Even though the field should not be added to the ChatMember structure, the accepted commit still does it https://github.com/tucnak/telebot/pull/407/commits/50ae1bb6b2bb4925affbd09dee8450cce3ea814b#diff-370ab673a241352de628b0935cae

Issue #404, opened 2021-05-23
API: support 5.x for v2


setval:
https://github.com/tucnak/telebot/pull/405
https://github.com/tucnak/telebot/pull/401

demget:
https://github.com/tucnak/telebot/commit/206507e3865e9e8cae3e428aaca683d15c64546e

setval:
https://github.com/tucnak/telebot/pull/407

setval:
https://github.com/tucnak/telebot/pull/408

PR #405, opened 2021-05-23
api: support 5.0

setval:
for v2

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -148,10 +148,14 @@ func (b *Bot) Ban(chat *Chat, member *ChatMember) error {
 }
 
 // Unban will unban user from chat, who would have thought eh?
-func (b *Bot) Unban(chat *Chat, user *User) error {
+func (b *Bot) Unban(chat *Chat, user *User, isBanned ...bool) error {
 	params := map[string]string{
 		"chat_id": chat.Recipient(),
 		"user_id": user.Recipient(),
+	}
+
+	if len(isBanned) > 0 {
+		params["only_if_banned"] = strconv.FormatBool(isBanned[0])
 	}
 
 	_, err := b.Raw("unbanChatMember", params)
--- a/admin.go
+++ b/admin.go
@@ -199,8 +199,9 @@ func (b *Bot) Promote(chat *Chat, member *ChatMember) error {
 	prv := member.Rights
 
 	params := map[string]interface{}{
-		"chat_id": chat.Recipient(),
-		"user_id": member.User.Recipient(),
+		"chat_id":      chat.Recipient(),
+		"user_id":      member.User.Recipient(),
+		"is_anonymous": member.IsAnonymous,
 	}
 	embedRights(params, prv)
 
--- a/admin.go
+++ b/admin.go
@@ -148,14 +148,15 @@ func (b *Bot) Ban(chat *Chat, member *ChatMember) error {
 }
 
 // Unban will unban user from chat, who would have thought eh?
-func (b *Bot) Unban(chat *Chat, user *User, isBanned ...bool) error {
+// forBanned does nothing if the user is not banned.
+func (b *Bot) Unban(chat *Chat, user *User, forBanned ...bool) error {
 	params := map[string]string{
 		"chat_id": chat.Recipient(),
 		"user_id": user.Recipient(),
 	}
 
-	if len(isBanned) > 0 {
-		params["only_if_banned"] = strconv.FormatBool(isBanned[0])
+	if len(forBanned) > 0 {
+		params["only_if_banned"] = strconv.FormatBool(forBanned[0])
 	}
 
 	_, err := b.Raw("unbanChatMember", params)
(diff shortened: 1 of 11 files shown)


DiscoreMe:
https://github.com/tucnak/telebot/issues/404

webhook.go
```diff
@@ -49,6 +49,9 @@ type Webhook struct {
 	ErrorUnixtime  int64  `json:"last_error_date"`
 	ErrorMessage   string `json:"last_error_message"`
 
+	IpAddress          string `json:"ip_address"`
```
demget: I believe simply `IP` is enough here.

webhook.go
```diff
@@ -49,6 +49,9 @@ type Webhook struct {
...
+	DropPendingUpdates bool   `json:"drop_pending_updates"`
```
demget: As with the `IP` field it's better to simplify it to `DropUpdates`.

chat.go
```diff
@@ -50,6 +51,11 @@ type Chat struct {
 	SlowMode         int        `json:"slow_mode_delay,omitempty"`
 	StickerSet       string     `json:"sticker_set_name,omitempty"`
 	CanSetStickerSet bool       `json:"can_set_sticker_set,omitempty"`
+	LinkedChatID     int64      `json:"linked_chat_id,omitempty"`
+	ChatLocation     *struct {
```
demget: Let's make another struct for `ChatLocation`, down here in `chat.go`.

admin.go
```diff
@@ -148,12 +148,16 @@ func (b *Bot) Ban(chat *Chat, member *ChatMember) error {
 }
 
 // Unban will unban user from chat, who would have thought eh?
-func (b *Bot) Unban(chat *Chat, user *User) error {
+func (b *Bot) Unban(chat *Chat, user *User, isBanned ...bool) error {
```
demget: `isBanned` is not entirely correct here, I'd name it `forBanned` which clearly shows that with this parameter, the function is going to affect only banned users.

telebot.go
```diff
@@ -144,6 +144,11 @@ const (
 	//
 	// Handler: func(*Message)
 	OnVoiceChatPartecipantsInvited = "\avoice_chat_partecipants_invited"
+
+	// Will fire on ProximityAlertTriggered
+	//
+	// Handler: func(*Message)
+	OnProximityAlertTriggered = "\aproximity_alert_triggered"
```
demget: My proposal is to rename it to `OnProximityAlert` and similarly for other related structs or fields, as word _triggered_ is already implied when you handle events.

bot.go
```diff
@@ -1395,15 +1395,32 @@ func (b *Bot) Pin(msg Editable, options ...interface{}) error {
...
 	}
+	if len(messageID) > 0 {
+		params["message_id"] = strconv.Itoa(messageID[0])
+	}
 
 	_, err := b.Raw("unpinChatMessage", params)
 	return err
 }
 
+// Unpin unpins all messages in a supergroup or a channel.
```
demget: Documentation for the function copy-pasted, but wasn't edited for `UnpinAll`.

media.go
```diff
@@ -195,9 +195,24 @@ type Location struct {
 	// Longitude
 	Lng float32 `json:"longitude"`
 
+	// Horizontal Accuracy
+	HA *float32 `json:"horizontal_accuracy,omitempty"`
```
demget: In this particular case, it's honestly better not to abbreviate words, _horizontal accuracy_ isn't something that can be recognized by these two letters.

media.go
```diff
@@ -195,9 +195,24 @@ type Location struct {
...
+	Heading int `json:"heading,omitempty"`
+
 	// Period in seconds for which the location will be updated
 	// (see Live Locations, should be between 60 and 86400.)
 	LivePeriod int `json:"live_period,omitempty"`
+
+	ProximityAlertRadius int `json:"proximity_alert_radius,omitempty"`
+}
+
+// ProximityAlertTriggered object represents sent whenever
```
demget: I suppose "object represents" words should be removed here.

chat.go
```diff
@@ -78,9 +78,10 @@ func (c *Chat) Recipient() string {
 type ChatMember struct {
 	Rights
 
-	User  *User        `json:"user"`
-	Role  MemberStatus `json:"status"`
-	Title string       `json:"custom_title"`
+	User        *User        `json:"user"`
+	Role        MemberStatus `json:"status"`
+	Title       string       `json:"custom_title"`
+	IsAnonymous bool         `json:"is_anonymous"`
```
demget: As we named `is_anonymous` field `Anonymous` once, we should use this name for further similar fields to preserve consistency. https://github.com/tucnak/telebot/blob/62cfa356f7f6a2c9e0ad011c32f3df2c076a57b4/polls.go#L22

demget (changes requested):
Some small adjustments to be done, everything else is fine.

demget (approved):

demget:
Such a great improvement for Telebot! Thank you for your work!

PR #409, opened 2021-05-25
Fix chat member handler

JasonKhew96:


demget:
Good catch, thank you.

PR #417, opened 2021-07-04
feat(api): support of ChatMemberCountByID

link-duan:


rusq:
Why don't you use this function?
```
func (*tb.Bot).Len(chat *tb.Chat) (int, error)
```

PR #427, opened 2021-07-27
Combine chatMember funcs

Rostish:


demget:
I think `ChatMember` will be a better name for a function.

demget:
Thank you

PR #436, opened 2021-08-17
Fix CreateInviteLink;EditInviteLink;RevokeInviteLink functions. Add t…

shindakioku:
Fix CreateInviteLink & EditInviteLink & RevokeInviteLink responses. Telegram returns 
```json
{"ok":true,"result":{"invite_link":"","creator":{"id":0,"is_bot":true,"first_name":"","username":""},"is_primary":false,"is_revoked":false}}
```

Fix RevokeInviteLink param: link -> invite_link name

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -6,11 +6,6 @@ import (
 	"time"
 )
 
-// ChatInviteLinkResult object represents an invite for a chat
-type ChatInviteLinkResult struct {
-	Result ChatInviteLink `json:"result"`
-}
-
 // ChatInviteLink object represents an invite for a chat.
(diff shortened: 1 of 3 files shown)


demget:
A good fix, thanks, for tests especially. There's only one thing: let's not define `ChatInviteLinkResult` to stay consistent with the previous API. We were always inlining the result in an anonymous structure, like here:  https://github.com/tucnak/telebot/blob/62cfa356f7f6a2c9e0ad011c32f3df2c076a57b4/stickers.go#L74

shindakioku:
Sure. Thanks.

Issue #446, opened 2021-09-25
SetGroupPermissions not working at all


Enrico204:
`SetGroupPermissions` is not working at all. Every combination or `Rights` will end up with all permissions removed.

Digging into the code I discovered that the problem is in the JSON request, which is malformed: all rights should be under the key `permissions` in the JSON object.

Tested with `v2` and `v3` (latest versions): both of them have this issue

Enrico204:
Fixed in #447

PR #447, opened 2021-09-25
Fix SetGroupPermissions JSON request body

Enrico204:
This pull request applies a fix for the issue #446 .

The JSON body of the request should contain rights under the `permissions` key


demget:
Good catch.

PR #454, opened 2021-10-15
Fix for v3: nativeContext.Text not returning Text for Inline Query

### LINKED ISSUE #453: v3: nativeContext.Text does not return the text for Inline Query
Problem: When calling Text() during inline query, the return value is empty (`""`).
Expected: Calling Text() returns Query.Text.

Will submit PR shortly.

rusq:
Hey Team, please review this short PR.  It fixes #453 for v3.


demget:
`Text()` is for fetching effective messages text only, either it's in callback, edited, or channel post message. So use `Data()` to get a query text.

database64128:
Got bitten by this when upgrading from v2 to v3. This is a major footgun in my opinion. Please reconsider!

PR #487, opened 2022-02-06
rate limiter: add

### LINKED ISSUE #268: proposal: api: implement the rate limiter
What do you think about implementing a default balancer to avoid flood limits requests api?

For example all Raw calls are queued and make sure that only 30 req/s can be done

setval:
There was a need to limit the number of requests coming to the servers. In order not to catch the anti-flood, an implementation has been made that allows you to limit the number of outgoing requests by n times per second.

Closes https://github.com/tucnak/telebot/issues/268


demget:
Changed a base to the custom branch to add the changes and features. Thank you, @setval.

ysomad:
Is there any chance it will be released soon? Really missing this feature

PR #488, opened 2022-02-09
Group\Channel middleware support

NexonSU:


demget:
Maybe we can use `c.Recipient` here? It checks for a chat to return, otherwise returns a sender.

NexonSU:
> Maybe we can use `c.Recipient` here? It checks for a chat to return, otherwise returns a sender.

Recipient cannot be used, because we need compare Sender().ID and only then Chat().ID.

PR #491, opened 2022-02-26
Add album caption parameter to SendAlbum

treapster:
Even though we can set caption in individual inputtable items, in some applications(such as crossposting bots) you parse and construct  attachments separately from text. After you create a slice of Inputtables, real media items are wrapped in interface and you can't change their caption(because InputMedia() returns value, not a reference). So i added caption parameter to sendAlbum which is assigned to Caption field of first media before sending if caption is not empty. If we don't want to change function signature, we can also add GetCaption/SetCaption to Inputtable interface, so let's discuss

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -285,7 +285,7 @@ func (b *Bot) Send(to Recipient, what interface{}, opts ...interface{}) (*Messag
 
 // SendAlbum sends multiple instances of media as a single message.
 // From all existing options, it only supports tele.Silent.
-func (b *Bot) SendAlbum(to Recipient, a Album, opts ...interface{}) ([]Message, error) {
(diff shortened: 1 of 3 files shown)


treapster:
It seems like there is still no way to send album with a caption, so i'd appreciate some attention from maintainers.

kyoto44:
agree, very useful

leqord:
How can I send an album with caption if caption parameter is now absent?

demget:
It breaks backward compatibility. The best way for this major version I can think of is adding a `SetCaption` method for the `Album` type. The inconvenient part of it is adding another huge type switch.

```go
var a tele.Album
a = // fill the album
a.SetCaption("...")

b.SendAlbum(to, a)
```

PR #495, opened 2022-03-13
Enable many endpoints for one handler

ksusonic:
I think that support of several bot commands with same purpose is useful feature


treapster:
Why not just call Handle for every endpoint you want?

ksusonic:
Ok

PR #503, opened 2022-03-30
Do not overwrite inline parsemode

LightningTipBot:
telebot v3 overwrites the default parsemode `ModeDefault ` for inline results if it differs from the bot's parse mode. This is not always desirable: inline article results to not support, for example, Markdown. 

This PR removes the code that overwrites the parse mode and hands the control over to the developer.


demget:
> inline article results to not support, for example, Markdown.

What do you mean? As far as I remember inline articles support any kind of formatting mode.

demget:
Could you please show a specific example where it can be essentially useful?

PR #504, opened 2022-04-03
webhook: add handler without http.Serve

robotomize:
I would like to be able to have a webhook handler that has no runtime. 
I have added an implementation of such a handler, it can be easily routed to an existing router without creating another listener.

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -161,6 +161,35 @@ func (h *Webhook) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.dest <- update
 }
 
+func NewWebhookHandler(webhook *Webhook) *WebhookHandler {
+	return &WebhookHandler{webhook: webhook}
+}
+
+// WebhookHandler is a simplified version of Webhook that has no separate runtime.
+//
+// It allows you to embed webhook into an existing router.
+type WebhookHandler struct {
+	webhook *Webhook
+}
+
+func (h *WebhookHandler) Poll(b *Bot, dest chan Update, stop chan struct{}) {
+	if err := b.SetWebhook(h.webhook); err != nil {
+		b.debug(err)
+		close(stop)
+		return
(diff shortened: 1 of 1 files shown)


Berezhnyk:
@robotomize Do you think this is going to help with running the webhook handler within the lambda function, which does not really allow you to listen to a port?

demget:
I like the idea but not the implementation. Wouldn't want to spawn another Webhook entity with a pretty similar logic. I'd better extend the existing implementation. I'll think about it.

robotomize:
@demget I agree, I sketched out a solution so as not to change anything in the current implementation and configuration. Cool if you make a better and more elegant solution!

suntong:
> Wouldn't want to spawn another Webhook entity with a pretty similar logic. I'd better extend the existing implementation.

Agree, no polling necessary. As comment in #536, for Webhook, as far as AWS lambda is concerned, what we need is merely to use `bot.HandleUpdate` to do the same internal stuff the library is doing from the webhook to *handle a single update*.

MetalBlueberry:
Can you do what is described [here](https://github.com/tucnak/telebot/pull/279)? It is working for me to deploy the code to AWS Lambda with just mapping the lambda request to the Update structure.

suntong:
> Can you do what is described [here](https://github.com/tucnak/telebot/pull/279)? It is working for me to deploy the code to AWS Lambda with just mapping the lambda request to the Update structure.

The provided example that goes with the PR is somehow removed later on (maybe the concern was that it was outdated after V3). Anyway, 

Would you be so kind as to give a working demo example for the current version please?

MetalBlueberry:
The example was removed, but It actually works.

In any case, here is a working example https://github.com/tucnak/telebot/pull/617

demget:
Since the provided example above makes more sense to me, closing this PR.

Issue #455, opened 2021-10-30
v3: recovering panics obstructs error searching


ema-pe:
Hi, I'm using `telebot.v3` module to build some Telegram bots. On my development environment I set to true the `Verbose` option to the bot, but when I run the bot and it panics it won't stop, instead it recovers the panics and prints an error message. This error message is not useful to understand where the bot panicked. I would suggest to add a new option that controls the deferred function inside [runHandler on utils.go](https://github.com/tucnak/telebot/blob/13d54ae7338ea472e64e84f42cacda591206bb3d/util.go#L30). I didn't open a pull request because there are multiple ways to solve this, maybe a discussion is needed.

dzen-it:
I faced into a problem too. I think it's not the best idea to suppress panic from business logic, since it takes a lot of debugging time to guess that the problem is not visible due to the library.

@tucnak It's great if you could make panic forwarding more explicit or exclude "recover".

Qusic:
My proposal is to simply add an option #506

demget:
`b.Use(middleware.Recover())` will be the new option to enable recover

PR #506, opened 2022-04-12
Add option NoRecover to make debugging the library easier

Qusic:
When `recover` is used, only the error message is printed. The debugger cannot catch the panic and it's hard to tell what is going wrong.


Qusic:
closing in favor of https://github.com/tucnak/telebot/issues/455#issuecomment-1257149518

Issue #288, opened 2020-05-11
Flow


tucnak:
### Introduction
There are multiple missing pieces from the current Telebot implementation.

One of these pieces is a feature, known as the context. I don't believe that it should be included _in_ the framework itself, as the state machine and the logic associated with it—is largely dependant on the implemention, therefore it must be implemented in the user code.

That said, we repeatedly end up implementing the context one way, or another. From the perspective of the code, it's trivial to do so. What is much more complicated is implementing complex multi-step interactions. This requires a large number of states, which adds to unnecessary verbosity and complicates the logic even further. To make things easier, I propose to introduce the concept of the _flow_.

### What is the flow?
The flow is simply a high-level algorithm, a series of actions and branches, which naturally occur in the logic of any sophisticated bot.

Please, consider the following algorithm:

1. `/start` A greeting message is shown, accompanied with a list of commands and an inline keyboard with key actions.
    - At this point, the state is empty and there is no current interaction.
    - One of the many stateful interactions may be triggered.
2. Either a command, or an inline button—invokes an interaction.
     - Generally, it doesn't matter which one it was, what matters is the flow.
     - The user state is now set to this particular interaction.
3. The bot requires a piece of media.
     - Depending on whether if it's a photo, a video or an audio recording, this step should leave to three further interactions.
     - Important: to remember the path of the flow, i.e. all the previous steps, including the outer interaction.
     - The user might want to go back and choose a different file without cancelling the whole interaction. This can be done by deleting the sent media, or simply invoking a _back_ action.
4. One of the three interactions get invoked, when a series of data is requested (a form.)
     - Each step of the way, data is harvested, validated separetely, and once form is validated as the whole.
     - At any given point, user is ought to be able to either go back in the form, or go _back_ to the previous interaction.
    - Important: to note that the form itself is a distinct iteration with its own state.
5. Once the interaction is complete, the user can be brought back to a certain _waiting state_, which may or may (step 1) not be completely stateless.
    - Depending on the particular waiting state, a different set of interactions may be available.

In order to implement the aforementioned algorithm, currently you would have to create a state machine of your own, and laboriously spell out each and every state, alongside with the numerous transition rules. Principally speaking, this is trivial, but as the interactions require multiple kinds of media and have many intermittent requirements, the implementation would have to be spread out across different handlers. The code will quickly grow uncontrollably.

### Proposal
[...]

PR #526, opened 2022-05-26
FSM support

bezumnui:
Finite state machine (like in aiogram)
In-memory based states storage.
States - the very important thing, that help you with logistic. But you may to know, that you will lost all states after restarting. (in Read.me more)


tucnak:
See [#288](https://github.com/tucnak/telebot/issues/288#issuecomment-1196262034) for clues. 

If we were to accept a FSM implementation, it would need to adhere at least in part to the model outlined there; think package `github.com/tucnak/telebot/v3/fsm` with the principle implementation of the `Interaction` interface.

PR #532, opened 2022-06-24
do cancel poll immediately on bot stop

realitycheck:
This PR proposes following:  

1. Cancel Raw() call immediately without waiting for the request's timeout when bot is about to stop.
This may become important if doing long polling with long timeout.
2. Make Bot.debug to be aware of Bot.verbose value to be more secure with sensible error messages (e.g. polling failures) which contains bot's token in url value.

### DIFF
--- a/api.go
+++ b/api.go
@@ -2,6 +2,7 @@ package telebot
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


tucnak:
This is a good change. I like it, especially the approach to URL concealment.

PR #539, opened 2022-07-30
support socks5 proxy for dial telebot with proxy to telegram api server

ja7ad:
Telegram API service is blocked by some providers, so you need to use a proxy or a VPN in order to connect to Telegram.
```go
package main

import (
	"gopkg.in/telebot.v3"
	"log"
	"time"
)

func main() {
	bot, err := telebot.NewBot(telebot.Settings{
		Token:  "token",
		Poller: &telebot.LongPoller{Timeout: 10 * time.Second},
		Proxy:  &telebot.Proxy{Address: "ip:port"},
	})
	if err != nil {
		log.Fatal(err)
	}

	bot.Handle("/hello", func(c telebot.Context) error {
		return c.Send("Hello!")
	})

	bot.Start()
}
```


demget:
I'm not sure about adding additional HTTP client-related logic, you can still use the Settings `Client` field for specifying the client with a custom transport dialer.

PR #543, opened 2022-08-18
Add secret_token support to webhook

zry98:
According to the Bot API 6.1 changes on June 20, 2022, a new optional parameter `secret_token` was added to the `setWebhook` method, if specified, webhook requests sent by Telegram will contain that value in the header `X-Telegram-Bot-Api-Secret-Token` ([documentation](https://core.telegram.org/bots/api#setwebhook:~:text=all%20pending%20updates-,secret_token,-String)).

I added this parameter to telebot's `Webhook` config, if configured, it will be used by `bot.SetWebhook`  and the webhook Poller (HTTP handler) will check it in incoming requests.


demget:
Thank you!

Issue #544, opened 2022-08-19
Template function "text" 


Googlom:
Trying to make bot reply with buttons with localized texts on them. From [here](https://github.com/tucnak/telebot/blob/c6bead11749b7621c8338ced665de9ef3028c48f/layout/layout.go#L516) I see that I can use `text` template function and it will load localized text using `lt.TextLocale()` from corresponding locales file. So in my layout.yml:
```yaml
buttons:
  testButton: Foo {{ text "bar" }}
```
In `en.yml`:
```yaml
bar: Bar
```
In handler:
```go
	btn := lt.Button(c, "testButton")
```
And it doesn't work - text in button is "Foo " but expected "Foo Bar".

But if I change map key in [layout.go](https://github.com/tucnak/telebot/blob/c6bead11749b7621c8338ced665de9ef3028c48f/layout/layout.go#L516) from `text` to, lets say, `text123` and in my layout.yml do the same:
```yaml
buttons:
  testButton: Foo {{ text123 "bar" }}
```
everything works as expected and text in button is "Foo Bar". Same for other locales.

So I'm not sure if it's telebots bug or I'm doing something wrong. Thanks in advance!

PR #546, opened 2022-08-22
fix the order in which template Funcs are applied

Googlom:
Solution to https://github.com/tucnak/telebot/issues/544.

The bug occured because `layout.template` methods result...
https://github.com/tucnak/telebot/blob/c6bead11749b7621c8338ced665de9ef3028c48f/layout/layout.go#L511-L520
was overwritten by method `template.Funcs(lt.funcs)` which applies...
https://github.com/tucnak/telebot/blob/c6bead11749b7621c8338ced665de9ef3028c48f/layout/layout.go#L285-L289
`builtinFuncs`:
https://github.com/tucnak/telebot/blob/c6bead11749b7621c8338ced665de9ef3028c48f/layout/layout.go#L99-L104

### DIFF
--- a/layout/layout.go
+++ b/layout/layout.go
@@ -282,7 +282,7 @@ func (lt *Layout) ButtonLocale(locale, k string, args ...interface{}) *tele.Btn 
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


demget:
Thank you, a good catch!

PR #550, opened 2022-08-30
Localized bot commands descriptions

Googlom:
implement `lt.CommandsLocale` methods to be able to set bot command desriptions for different locales


demget:
A pretty good way to manage localized commands via `telebot/layout`! Thanks.

PR #567, opened 2022-11-10
Adds ability to send a photo with entities

rieneuf:
Entities to be represented as a JSON string.


demget:
This is not right, entities are the sending option in telebot, simply pass them right after the photo object.

```go
b.Send(to, &tele.Photo{}, entities, tele.Silent, ...)
```

PR #568, opened 2022-11-12
bot: use Recipient instead of Chat in Leave and Unpin methods

opmega:
These methods used Chat when they only call the Recipient() from the Recipient interface.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -953,7 +953,7 @@ func (b *Bot) StopPoll(msg Editable, opts ...interface{}) (*Poll, error) {
 }
 
 // Leave makes bot leave a group, supergroup or channel.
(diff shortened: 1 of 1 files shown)


opmega:
Having Recipient as parameter allows usage of `telebot.ChatID(id)`

demget:
Yeah, the thing is we don't want to expect `Sender` to be used in these calls. But the `ChatID` argument will be a more useful case anyway, and passing a proper chat recipient will be up to the user then. So I think we'll merge it.

demget:
Are there other methods we need to change for consistency?

opmega:
When I did this change these 3 methods were the only ones I found but it has been a while. If necessary I could go ahead and verify.

PR #570, opened 2022-11-21
errors: add join request related errors

awohsen:
I've been creating a bot to accept join requests for high request rate channels of mine.

These were the errors that I wished to handle but I didn't find them at `errors.go` so I added them.


awohsen:
I just found another one which I have no idea what is the meaning of it:
`telegram: Bad Request: HIDE_REQUESTER_MISSING (400)`

also if the user never requested to join:
`telegram: Bad Request: USER_ID_INVALID (400)`

PR #579, opened 2023-02-05
media: add missed CustomEmoji field to the Sticker

vlad-lukyanov:
### Bot API 6.2
**Custom Emoji Support**

- Added the fields type and custom_emoji_id to the class Sticker.


demget:
Please remove the `,omitempty` part from the `json` tag. This field is not used in the requests.

PR #588, opened 2023-04-03
Fix: handler middleware is overwritten when combined with group or global middleware

mkuznets:
## Summary

When handler-scoped middlewares are combined with global- or group-scoped middlewares, the concatenation is now done using a newly allocated slice instead of `append()` to the existing slice.

## Motivation

Here's how handler-scoped middleware was initialised before:

```go
// m is a slice of handler-scoped middleware

if len(b.group.middleware) > 0 {
	m = append(b.group.middleware, m...)
}

handler := func(c Context) error {
	return applyMiddleware(h, m...)(c)
}
```

If `b.group.middleware` has extra capacity to fit the elements from `m`, the `append` will place the elements of `m` right next to the elements of `b.group.middleware` without allocating new memory. The expanded slice will be captured by the handler.

However, if we add another handler with a middleware, the `append` will use the same extra capacity *again*, overwriting the handler-scoped middleware added before.

The same issue applies when group-scoped middleware is combined with the global one.

Example:

```go
bot, _ := telebot.NewBot(telebot.Settings{Offline: true, Synchronous: true})
// At least three middlewares to trigger extra capacity in the middleware slice
bot.Use(middleware.AutoRespond())
bot.Use(middleware.IgnoreVia())
bot.Use(middleware.Logger(log.Default()))

bot.Handle("/secret", func(c telebot.Context) error { panic("unauthorised access") }, middleware.Whitelist(42))

// middleware.Blacklist clobbers middleware.Whitelist from the previous handler
bot.Handle("/public", func(c telebot.Context) error { return c.Send("Hello, world!") }, middleware.Blacklist())

u := telebot.Update{
	Message: &telebot.Message{
		Text:   "/secret",
		Sender: &telebot.User{ID: 1337},
	},
}

bot.ProcessUpdate(u) // panic: unauthorised access
```

## Tests

- Added two tests that demonstrate the issue
- Added an extra test to verify the call order of middlewares. This is to ensure that my changes did not affect anything.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -173,12 +173,15 @@ var (
 //		b.Handle("/ban", onBan, middleware.Whitelist(ids...))
 //
 func (b *Bot) Handle(endpoint interface{}, h HandlerFunc, m ...MiddlewareFunc) {
+	mw := m
 	if len(b.group.middleware) > 0 {
-		m = append(b.group.middleware, m...)
+		mw = make([]MiddlewareFunc, 0, len(b.group.middleware)+len(m))
+		mw = append(mw, b.group.middleware...)
+		mw = append(mw, m...)
 	}
 
 	handler := func(c Context) error {
-		return applyMiddleware(h, m...)(c)
+		return applyMiddleware(h, mw...)(c)
 	}
 
 	switch end := endpoint.(type) {
--- a/bot_test.go
+++ b/bot_test.go
@@ -373,6 +373,109 @@ func TestBotOnError(t *testing.T) {
 	assert.True(t, ok)
 }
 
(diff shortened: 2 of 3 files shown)


roreng:
Encountered a similar issue. 
Thank you, @mkuznets, for providing the solution. 

@demget @tucnak, could you please review this pull request?

demget:
Such a great work, @mkuznets. Thank you and sorry for the long wait.

Cases like this show how knowing the language internals is significant, as well as having qualitative tests...

I moved the append-thing to the separate function and prettied the tests, check it out: #628

PR #589, opened 2023-04-06
MarshalJSON should produce unique result ID

MikhailChe:
Result variable inside for-loop actually have the same address on each iteration, so we need to take address in underlying slice.


demget:
@MikhailChe why did you close this PR?

MikhailChe:
> @MikhailChe why did you close this PR?

I guess I was just cleaning up something and closed it by accident.

demget:
Thank you

PR #597, opened 2023-04-24
Chore: replaced `io/ioutil` package

AH-dark:
I replaced `io/ioutil` package with `os` and `io` package.

With the two proposals accepted (https://github.com/golang/go/issues/42026 and https://github.com/golang/go/issues/40025), the package io/ioutil will be deprecated and new code is encouraged to use the respective implementations in the packages io and os.

Because the Go version used by GitHub Action's workflow is already 1.16, I also increased the version of gomod to 1.16.

### DIFF
--- a/api.go
+++ b/api.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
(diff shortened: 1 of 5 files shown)


demget:
Good job, thank you!

PR #610, opened 2023-07-20
Allow initialisation from io.Reader

rusq:
## Problem

layout package expects the yaml file to be available on disk.  With embedding,
it could be convenient to store the yaml on an embedded filesystem, or embed as
a string at compile-time.

## Proposed Solution

- Implement `FromReader` and `ReadDefault` that allow to read the layout from
  any reader.
- Modify New and NewDefault to use FromReader and ReadDefault correspondingly.

This allows to use the embedded layout.yaml file, i.e.:
```go
//go:embed layout.yaml
var layoutYAML string

func foo() {
    //...
    l, err := layout.FromReader(strings.NewReader(layoutYAML))
    if err != nil {
        // handle err
    }
    // use layout
    // ...
}
```

### DIFF
--- a/layout/layout.go
+++ b/layout/layout.go
@@ -3,8 +3,9 @@ package layout
 import (
 	"bytes"
 	"encoding/json"
-	"io/ioutil"
+	"io"
 	"log"
+	"os"
 	"strings"
 	"sync"
 	"text/template"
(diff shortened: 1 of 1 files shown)


rusq:
As I experiment with layout, I think it's better to add the support for `fs.FS`.  I'll create a PR when I have something, experimenting with layout ATM.  Would be interesting to know your thoughts @demget .

PR #611, opened 2023-08-06
Added support for telegram local API

ya-meh:
Local API server is necessary, when working with files bigger than 20 MB, increasing threshold for files up to 2 GB.

Only downloading files needs revision, uploading works out of the box.

https://github.com/tdlib/telegram-bot-api


demget:
Manage the file location logic on your side, you still have the `FilePath` pointing to it, right?

PR #558, opened 2022-10-11
bot: add support for the test environment

caalberts:
This adds support for Telegram test environment, as requested in #556.

https://core.telegram.org/bots/webapps#using-bots-in-the-test-environment

PR #613, opened 2023-08-20
bot: add an ability to use custom url builder

d0ctr:
Same reasons as for #558, but a more robust solution as allows to use one of the setting's fields instead of two (URL and test env switcher)

### DIFF
--- a/api.go
+++ b/api.go
@@ -20,7 +20,7 @@ import (
 // It also handles API errors, so you only need to unwrap
 // result field from json data.
 func (b *Bot) Raw(method string, payload interface{}) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


demget:
I don't like the `URLBuilder` name, and what's the use case of being able to pass it to the bot settings?

d0ctr:
> I don't like the `URLBuilder` name, and what's the use case of being able to pass it to the bot settings?

Telegram servers have a test environment, that has a different path: `<base>/bot<token>/test/<method>`.
The name can be whatever you want, but without the builder or a special option the test environment is inaccessible.

demget:
For testing purposes, I'll merge #558.

PR #625, opened 2023-10-25
api: fix race for a stopClient channel

odourain:
# Problem

Before starting the bot I do a setCommands and its goroutine [`which reads channel variable`] does not have time to complete before the start method [`which writes to channel variable`].

Previous read at 0x00c0015a6628 by goroutine 106:
 [ https://github.com/tucnak/telebot/blob/v3/api.go#L39](https://github.com/tucnak/telebot/blob/v3/api.go#L39)

Write at 0x00c0015a6628 by main goroutine:
 [ https://github.com/tucnak/telebot/blob/v3/bot.go#L205](https://github.com/tucnak/telebot/blob/v3/bot.go#L205)

# Decision

Use mutex for use channel `stopClient`.

Also added changes from https://github.com/caalberts/telebot to work with testnet.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -49,6 +49,7 @@ func NewBot(pref Settings) (*Bot, error) {
 		synchronous: pref.Synchronous,
 		verbose:     pref.Verbose,
 		parseMode:   pref.ParseMode,
+		useTestEnv:  pref.UseTestEnv,
 		client:      client,
 	}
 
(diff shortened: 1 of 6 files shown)


DmytroSolodukhin (approved):

bot.go
```diff
@@ -198,10 +207,11 @@ func (b *Bot) Start() {
 	}
 
 	// do nothing if called twice
-	if b.stopClient != nil {
+	if b.getChanStopClient() != nil {
```
davseby, line 210: We're still racing here, it would be better to encapsulate nil checking and creating into a one mutex call. Same with the stop.

davseby (commented):
Before I noticed this PR, I did a fix of my own. I leave it up for the maintainers to decide which one to use, however I did notice some issues with this one 🤷🏿.

https://github.com/tucnak/telebot/pull/641

Issue #584, opened 2023-03-13
webhook: panic: close of closed channel


archimed-shaman:
`gopkg.in/telebot.v3 v3.1.3`

Error:
```
panic: close of closed channel

goroutine 49 [running]:
gopkg.in/telebot%2ev3.(*Webhook).waitForStop(...)
	/go/pkg/mod/gopkg.in/telebot.v3@v3.1.3/webhook.go:155
gopkg.in/telebot%2ev3.(*Webhook).Poll.func1(0x0?)
	/go/pkg/mod/gopkg.in/telebot.v3@v3.1.3/webhook.go:142 +0x34
created by gopkg.in/telebot%2ev3.(*Webhook).Poll
	/go/pkg/mod/gopkg.in/telebot.v3@v3.1.3/webhook.go:141 +0x1d9
```

Looks like this happens because `stop` channel is closed in two places:
https://github.com/tucnak/telebot/blob/fbd35f210343604adb89e96859a9639d9366e966/bot.go#L222
https://github.com/tucnak/telebot/blob/fbd35f210343604adb89e96859a9639d9366e966/webhook.go#L123

suntong:
Hi @archimed-shaman, would you share how you get this far please? I'm having troubles making webhook work for tucnak/telebot. Moreover, webhook does not need poller, right?

ayasechan:
The reason for this problem is
If you do not configure the Listen field
poll function will directly call the waitForStop method to close the channel

When you call the Stop function, this function will call waitForStop for the second time closing the channel to cause panic

https://github.com/tucnak/telebot/blob/71ac2995cc5d7379c35fceb3b64ce2c693e79e0c/webhook.go#L131-L133

Dementor316:
Am I correct in understanding that for an elegant solution to the problem, we simply do not need to call the Stop() method when terminating the bot process? If I don't call this method, no panic occurs, but how necessary is its invocation and how much can the absence of calling the Stop() method when using WebHooks lead to negative consequences? I would be grateful if someone could help me understand this issue.

demget:
#641

tarik02:
It was not actually fixed, If you call `Stop`, you close the channel. And then, in `waitForStop` you cancel it the next time...

kyrylo:
I confirm this is not fixed.

Please, reopen (the stacktrace is identical to the OP's stacktrace).

PR #641, opened 2023-12-28
bot,api: fix stop channel race

davseby:
# Problem
Running tests with race flag, you may get these errors:

Previous read at 0x00c0015a6628 by goroutine 106:
https://github.com/tucnak/telebot/blob/v3/api.go#L39

Write at 0x00c0015a6628 by main goroutine:
https://github.com/tucnak/telebot/blob/v3/bot.go#L205

# Solution
Use a mutex to manage stopClient channel.

Note: Kinda a duplicate of https://github.com/tucnak/telebot/pull/625, however it ensures that multiple stops or starts doesn't cause race issues.

### DIFF
--- a/api.go
+++ b/api.go
@@ -27,18 +27,21 @@ func (b *Bot) Raw(method string, payload interface{}) ([]byte, error) {
 		return nil, err
 	}
 
-	// Cancel the request immediately without waiting for the timeout  when bot is about to stop.
+	// Cancel the request immediately without waiting for the timeout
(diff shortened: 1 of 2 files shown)


demget:
Looks much better than #625, thank you. This is going to be merged.

demget:
@davseby am I right that your changes should fix #584?

davseby:
> @davseby am I right that your changes should fix #584?

Hi @demget, yes, it's supposed to.

PR #647, opened 2024-02-03
callback: add missing struct fields

rapour:
The telegram Bot API includes `chat_instance` and `game_short_name` in its [CallbackQuery object](https://core.telegram.org/bots/api#callbackquery). Particularly, the `game_short_name` is currently missing in the `Callback` object of the update when a user requests a game's url from the bot and it's also not deserialized in the `Game` object of the callback's message. 

I think it might be nice to have the Callback struct conforming with Telegram's CallbackQuery object and also a lot easier to implement than somehow populating the Game struct in the [message](https://github.com/tucnak/telebot/blob/71ac2995cc5d7379c35fceb3b64ce2c693e79e0c/message.go#L134).

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -26,13 +26,13 @@ type Callback struct {
 	// a bad client can send arbitrary data in this field.
 	Data string `json:"data"`
 
-	// Global identifier, uniquely corresponding to the chat to
-	// which the message with the callback button was sent.
+	// ChatInstance is a global identifier, uniquely corresponding to
+	// the chat to which the message with the callback button was sent.
 	ChatInstance string `json:"chat_instance"`
 
-	// Unique identifier of the game for which a URL is requested
-	// from the bot when a user presses the Play button of that game.
-	// GameShortName may be empty
+	// GameShortName is a unique identifier of the game for which a URL
+	// is requested from the bot when a user presses the Play button of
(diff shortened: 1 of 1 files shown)


callback.go
```diff
@@ -26,6 +26,15 @@ type Callback struct {
 	// a bad client can send arbitrary data in this field.
 	Data string `json:"data"`
 
+	// Global identifier, uniquely corresponding to the chat to
```
demget: Please fix the documentation:

```suggestion
	// ChatInstance is a global identifier, uniquely corresponding to the chat to
```

callback.go
```diff
@@ -26,6 +26,15 @@ type Callback struct {
...
+	// which the message with the callback button was sent.
+	ChatInstance string `json:"chat_instance"`
+
+	// Unique identifier of the game for which a URL is requested
```
demget: Same here.

demget (changes requested):

callback.go
rapour: Done

callback.go
rapour: Done

demget:
Good catch, thank you for filling the hole in our API.

PR #648, opened 2024-02-07
inline: add game result support

rapour:
[Game](https://core.telegram.org/bots/api#inlinequeryresultgame) is one of the possible types a bot can send as inline query result.

### DIFF
--- a/inline_types.go
+++ b/inline_types.go
@@ -60,12 +60,14 @@ func (r *ResultBase) Process(b *Bot) {
 	}
 }
 
-// GameResult represents a game.
+// GameResult represents a game. Game is a content type
+// supported by Telegram, which can be sent back to the
+// user as a result for an inline query.
 type GameResult struct {
 	ResultBase
 
-	// Unique identifier of the game.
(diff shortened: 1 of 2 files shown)


inline_types.go
```diff
@@ -60,6 +60,14 @@ func (r *ResultBase) Process(b *Bot) {
 	}
 }
 
+// GameResult represents a game.
+type GameResult struct {
+	ResultBase
+
+	// Unique identifier of the game.
```
demget: Documentation

inline_types.go
```diff
@@ -60,6 +60,14 @@ func (r *ResultBase) Process(b *Bot) {
...
+	GameShortName string `json:"game_short_name"`
```
demget: `ShortName`

demget (changes requested):

inline_types.go
rapour: Done

inline_types.go
rapour: Done

Issue #629, opened 2023-11-07
telebot: add AllUpdates flag for the long poller settings


demget:
A convenient way to specify all the updates the bot will receive. To do this you need to specify all the update kinds possible to the `allowed_updates` param.

The list of updates:
```
message
edited_message
channel_post
edited_channel_post
inline_query
chosen_inline_result
callback_query
shipping_query
pre_checkout_query
poll
poll_answer
my_chat_member
chat_member
chat_join_request
```

PR #661, opened 2024-03-01
media: implement SetCaption in Album

### LINKED ISSUE #631: proposal: add a way to set the caption for Album
The problem is described here: #491 & https://github.com/tucnak/telebot/pull/491#issuecomment-1817536362.

nentenpizza:
Closes #631

Also #629

### DIFF
--- a/media_test.go
+++ b/media_test.go
@@ -1,0 +1,43 @@
+package telebot
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestAlbumSetCaption(t *testing.T) {
+	var a Album
+	a = append(a, &Photo{Caption: "wrong_caption"})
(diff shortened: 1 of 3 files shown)


media_test.go
```diff
@@ -0,0 +1,43 @@
+package telebot
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestAlbumSetCaption(t *testing.T) {
```
demget, line 8: Tests could be unified:
```go
tests := []struct{
  name  string
  media Media
}{
  {
    name:  "photo",
    media: &Photo{Caption: "..."},
  },
  {
    // ...
  }
}
for _, tt := range tests {
  t.Run(tt.name, func(t *testing.T) {
    // ...
  })
}
```

demget (changes requested):

PR #681, opened 2024-04-13
FIX: keep order of commands, better docs

dvordrova:
Came across some difficulties when used layout, one of that is commands usage:
- every restart I have different set of commands, owning to it is map (made it list)
- not obvious to find way of templating commands (added to example.yml) and tests
- not obvious how to use SetCommands for all not specified locales (added it in layout docstring)

### DIFF
--- a/layout/layout_test.go
+++ b/layout/layout_test.go
@@ -32,13 +32,26 @@ func TestLayout(t *testing.T) {
 	assert.Equal(t, &tele.LongPoller{}, pref.Poller)
 	assert.Equal(t, pref, ltfs.Settings())
 
-	assert.ElementsMatch(t, []tele.Command{{
+	assert.Equal(t, []tele.Command{{
 		Text:        "start",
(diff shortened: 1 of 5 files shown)


layout/layout_test.go
```diff
@@ -32,13 +32,26 @@ func TestLayout(t *testing.T) {
...
+	assert.Equal(t, []tele.Command{{
 		Text:        "start",
 		Description: "Start the bot",
 	}, {
 		Text:        "help",
 		Description: "How to use the bot",
+	}, {
+		Text:        "settings",
+		Description: "{{ text `cmd_settings` }}",
 	}}, lt.Commands())
```
demget, line 44: Why to keep `lt.Commands()` then if they're not executed as a template?

demget (changes requested):

demget:
Thank you @dvordrova, please rebase from `v3.4` and fix the merge conflicts

PR #686, opened 2024-05-24
context: add boost send cases, boost: fix UserBoosts

Nash-Well:


context.go
```diff
@@ -301,6 +301,20 @@ func (c *nativeContext) Sender() *User {
...
+	case c.u.BoostRemoved != nil:
+		src := c.u.BoostRemoved.Source
+		if src == nil {
+			return nil
+		}
+
+		return src.Booster
 	default:
 		return nil
 	}
```
demget:
```suggestion
	case c.u.Boost != nil:
		if b := c.u.Boost.Boost; b != nil && b.Source != nil {
			return b.Boost.Source.Booster
		}
	case c.u.BoostRemoved != nil:
		if  b := c.u.BoostRemoved; b.Source != nil {
			return src.Booster
		}
	}
	return nil
```

demget (changes requested):

PR #697, opened 2024-06-17
Added forward handler

### LINKED ISSUE #696: Cant catch forwarded messages
If i forward message to bot, I cant catch it. Message text is "" and filled fields are Origin, Caption and some others.

nikepan:
Resolve #696

### DIFF
--- a/update.go
+++ b/update.go
@@ -74,6 +74,10 @@ func (b *Bot) ProcessUpdate(u Update) {
 			return
 		}
 
+		if m.Origin != nil && m.Text == "" {
+			b.handle(OnForward, c)
+		}
+
 		if m.Contact != nil {
 			b.handle(OnContact, c)
 			return
--- a/update.go
+++ b/update.go
@@ -66,6 +66,10 @@ func (b *Bot) ProcessUpdate(u Update) {
 				return
 			}
 
+			if m.ReplyTo != nil {
+				b.handle(OnReply, c)
+			}
+
 			b.handle(OnText, c)
 			return
 		}
@@ -74,7 +78,7 @@ func (b *Bot) ProcessUpdate(u Update) {
 			return
 		}
 
-		if m.Origin != nil && m.Text == "" {
+		if m.Origin != nil {
 			b.handle(OnForward, c)
 		}
 
--- a/telebot.go
+++ b/telebot.go
@@ -46,6 +46,7 @@ const (
 	// Basic message handlers.
 	OnText                 = "\atext"
 	OnForward              = "\aforward"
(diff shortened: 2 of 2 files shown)


update.go
```diff
@@ -74,6 +74,10 @@ func (b *Bot) ProcessUpdate(u Update) {
 			return
 		}
 
+		if m.Origin != nil && m.Text == "" {
```
demget: Why is the text should be empty?

demget (changes requested):

update.go
nikepan: It seems to be always empty when forwarding. but if it is for example a reply, the text will not be empty.

demget: The emptiness of `m.Text` is already checked here

https://github.com/tucnak/telebot/blob/350ac26ac84216662c19baea128c18fecbc0eb84/update.go#L43

demget: Could you please also add the same event for reply to a message? There is a `ReplyTo` field in the `Message`

demget (changes requested):

nikepan:
Here ok? https://github.com/tucnak/telebot/blob/350ac26ac84216662c19baea128c18fecbc0eb84/update.go#L68

demget:
> Here ok?
> 
> https://github.com/tucnak/telebot/blob/350ac26ac84216662c19baea128c18fecbc0eb84/update.go#L68

Yes

nikepan:
I made some tests again. text="" if it forward from channel. if it from chat, text is original and trigger text events.
I tested reply event. It works fine.

demget:
> I made some tests again. text="" if it forward from channel. if it from chat, text is original and trigger text events. I tested reply event. It works fine.

Then you should move the `m.Origin` check before the `m.Text` check. Complete this and I'll merge your PR into `v3.4`.

nikepan:
I think this action will create a backwards incompatibility for applications that recognise forwarding from chats now as a normal message

demget:
> I think this action will create a backwards incompatibility for applications that recognise forwarding from chats now as a normal message

Oh yeah, you're definitely right. We should check somehow if there is `OnForward` or `OnReply` handler registered, before trying to execute them, if not, the bot should follow the usual `OnText` flow.

Do you think you can implement it in a separate PR?

nikepan:
I don't think that's a good idea. Because it violates the text filters.  I took another look. The current version doesn't break anything if there is no OnForward handler. But if there is one, 2 events are triggered. I think the most correct thing would be to move the check down. Then it will be possible to catch the event if text filters are not triggered, which I believe should be a priority.

PR #698, opened 2024-06-17
BotAPI7.4

Apolisk:


bot.go
```diff
@@ -1280,3 +1280,18 @@ func extractEndpoint(endpoint interface{}) string {
 	}
 	return ""
 }
+
+// SendStarRefund returns a successful payment in Telegram Stars.
+func (b *Bot) SendStarRefund(to Recipient, chargeID string) error {
```
demget, line 1285:
```suggestion
func (b *Bot) RefundStars(to Recipient, chargeID string) error {
```

demget:
- Check for fields (that are not required for Telegram Stars) presence here (https://github.com/tucnak/telebot/blob/a8acf843014bee762a8426a67d3b61ea4830200a/payments.go#L97)

demget (changes requested):

PR #699, opened 2024-06-18
BotAPI7.3

Apolisk:


demget:
- Create an alias for `math.MaxInt32` name it `LiveForever` (live_period)
- Remove `ChatFullInfo`, and include new fields in the `Chat` struct

PR #702, opened 2024-06-23
Comprehensive Update: Implement Webhook Options, Payment Support, and Various Enhancements

pridclow68:
1. add UseWebhook and NoWebhook options to SendOptions
2. update regex pattern for cbackRx in bot.go
3. add OnForward constant to telebot.go and handle it in ProcessUpdate function
4. handle unauthorized status code in extractOk function
5. add support for payments in Telegram Stars RefundStarPayment and GetStarTransactions

Examples of how to use the newly added code: https://github.com/pridclow68/telebot/blob/v3_example/example/tgbot.go


demget:
- What's the point of having webhook options?
- Why is the `UseWebhook` added to every context wrapper?
- We don't really want to use `reflect` inside telebot
- Overall the PR is pretty messy and should be divided into different proposals and discussions, and smaller PRs as well

PR #705, opened 2024-06-26
BUG: reactions count bug

Dionid:
On updates of post reactions I've got error: `json: cannot unmarshal array into Go struct field MessageReactionCount.Result.message_reaction_count.reactions of type telebot.ReactionCount]`

I've found, if you use "message_reaction" allowed_updates, that `message_reaction_count.reactions` came as list, not just one.

I can't be sure that this will be the same for all cases, but it is in mine.


demget:
Good catch @Dionid, thank you

PR #708, opened 2024-06-30
media: added field vcard to Contact

Apolisk:


media.go
```diff
@@ -328,6 +328,7 @@ type Contact struct {
 	// (Optional)
 	LastName string `json:"last_name"`
 	UserID   int64  `json:"user_id,omitempty"`
+	Vcard    string `json:"vcard"`
```
demget: VCard

demget (changes requested):

PR #711, opened 2024-07-05
Bot API 7.6

Apolisk:


context.go
```diff
@@ -169,6 +169,9 @@ type Context interface {
 
 	// Set saves data in the context.
 	Set(key string, val interface{})
+
+	// Payment returns payment instance.
+	Payment() *Payment
```
demget, line 174: Move it to other functions that return update/message types

context.go
```diff
@@ -538,3 +541,7 @@ func (c *nativeContext) Get(key string) interface{} {
 	defer c.lock.RUnlock()
 	return c.store[key]
 }
+
+func (c *nativeContext) Payment() *Payment {
```
demget, line 545: Here as well. Keep file ordered and easy to navigate.

media.go
```diff
@@ -34,6 +34,30 @@ type InputMedia struct {
 	HasSpoiler           bool     `json:"has_spoiler,omitempty"`
 }
 
+type InputPaidMedia struct {
```
demget, line 37: `InputMedia` has the same set of fields, so we can avoid creating another structure

media.go
```diff
@@ -34,6 +34,30 @@ type InputMedia struct {
...
+	Media             string `json:"media"`
+	Thumbnail         string `json:"thumbnail,omitempty"`
+	Width             int    `json:"width,omitempty"`
+	Height            int    `json:"height,omitempty"`
+	Duration          int    `json:"duration,omitempty"`
+	SupportsStreaming bool   `json:"supports_streaming"`
+}
+
+type PaidMediaInfo struct {
+	StarCount int         `json:"star_count"`
```
demget, line 48:
```suggestion
	Stars int         `json:"star_count"`
```

media.go
```diff
@@ -34,6 +34,30 @@ type InputMedia struct {
...
+	PaidMedia []PaidMedia `json:"paid_media"`
+}
+
+type PaidMedia struct {
+	Type     string      `json:"type"`
+	Photo    []photoSize `json:"photo,omitempty"`
```
demget, line 54:
```suggestion
	Photo    *Photo `json:"photo,omitempty"`
```

demget (changes requested):
- `sendPaidMedia` is not implemented

PR #720, opened 2024-07-23
feat: set bot default menu button

vika2603:
The original `Bot.SetMenuButton` method cannot set the default menu button, which solves the problem


Nurshinov:
I tested it now and everything works, thanks

vika2603:
@Nurshinov Did you also encounter the same problem?

Nurshinov:
> @Nurshinov Did you also encounter the same problem?

Yes, i did

PR #736, opened 2024-08-29
 Add structured unknown error from telegram api

ev-gusev:
There are cases when api telegram returns errors that do not have static text, for example: 
```
Bad Request: can't parse entities: Can't find end tag corresponding to start tag "b"
```

In this case, it is impossible to determine on the library user side that this error is a Bad Request, as it will be generic and will only contain the error text

This PR solves this problem and allows to handle such errors as follows:
```
	var e *telebot.Error
	if errors.As(err, &e) {
		if e.Code == 400 {
			...
		}
	}
```

Errors that were previously predefined as bad request will be handled in the same form

### DIFF
--- a/bot_raw.go
+++ b/bot_raw.go
@@ -330,14 +330,11 @@ func extractOk(data []byte) error {
 			err:        NewError(e.Code, e.Description),
 			RetryAfter: int(retryAfter.(float64)),
 		}
-	case http.StatusBadRequest:
+	default:
 		err = &Error{
 			Code:        e.Code,
 			Description: e.Description,
 		}
-
-	default:
-		err = fmt.Errorf("telegram: %s (%d)", e.Description, e.Code)
 	}
 
 	return err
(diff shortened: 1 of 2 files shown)


demget:
Hey @ev-gusev, good catch. I think we can change `default` case instead to generalize such conversion

ev-gusev:
> Hey @ev-gusev, good catch. I think we can change `default` case instead to generalize such conversion

I actually thought about it, but I thought that there might be library users who used regular expressions and parsed these errors that were not of type Error to accomplish the goals I was pursuing . So I didn't risk it

ev-gusev:
@demget What do you think about it?

demget:
@ev-gusev we will include this in the breaking V4 release

ev-gusev:
> @ev-gusev we will include this in the breaking V4 release

I got you, changed the default case

ev-gusev:
@demget Is something wrong?

PR #738, opened 2024-08-30
FIX: keep order of commands, better docs

dvordrova:
Came across some difficulties when used layout, one of that is commands usage:

every restart I have different set of commands, owning to it is map (made it list)
not obvious to find way of templating commands (added to example.yml) and tests
not obvious how to use SetCommands for all not specified locales (added it in layout docstring)

Previous try https://github.com/tucnak/telebot/pull/681

### DIFF
--- a/layout/layout_test.go
+++ b/layout/layout_test.go
@@ -32,13 +32,26 @@ func TestLayout(t *testing.T) {
 	assert.Equal(t, &tele.LongPoller{}, pref.Poller)
 	assert.Equal(t, pref, ltfs.Settings())
 
-	assert.ElementsMatch(t, []tele.Command{{
+	assert.Equal(t, []tele.Command{{
 		Text:        "start",
 		Description: "Start the bot",
 	}, {
 		Text:        "help",
 		Description: "How to use the bot",
+	}, {
+		Text:        "settings",
+		Description: "{{ text `cmd_settings` }}",
 	}}, lt.Commands())
+	assert.Equal(t, []tele.Command{{
+		Text:        "start",
+		Description: "Start the bot",
+	}, {
+		Text:        "help",
+		Description: "How to use the bot",
+	}, {
+		Text:        "settings",
+		Description: "Settings",
+	}}, lt.CommandsLocale("en"))
 
 	assert.Equal(t, "string", lt.String("str"))
 	assert.Equal(t, 123, lt.Int("num"))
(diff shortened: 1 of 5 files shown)


layout/layout_test.go
```diff
@@ -32,13 +32,26 @@ func TestLayout(t *testing.T) {
...
-	assert.ElementsMatch(t, []tele.Command{{
+	assert.Equal(t, []tele.Command{{
 		Text:        "start",
 		Description: "Start the bot",
 	}, {
 		Text:        "help",
 		Description: "How to use the bot",
+	}, {
+		Text:        "settings",
+		Description: "{{ text `cmd_settings` }}",
```
dvordrova, line 43: previous pr had a question about testing this part of template
https://github.com/tucnak/telebot/pull/681#discussion_r1711361203

The reasons to keep `lt.Commands` in tests:
1) make it clear what happens when you use templates with `lt.Commands()`
2) the test already contained testing of both Commands and CommandsLocale functions, kept it here

dvordrova: @demget should I check anything else before you run CI workflow?

demget:
@dvordrova I meant the fact that `Commands` is exposed but doesn't execute the commands as a template, which leads to the raw description content. What would I use it for?

https://github.com/tucnak/telebot/pull/681/files/d9dcbb41bfee07192e91c90fc9d4d3eba7b1ddfc#diff-36c59fae7f0af10b179a0e3781fe9ea9d14c5e5a99d6a36a1f24767010bca403R43

dvordrova:
> What would I use it for?

@demget so you wouldn't
it is just test representation of comment I added `// "commands" must be not templates`
https://github.com/tucnak/telebot/pull/738/files#diff-cf94c60ab5f782ca8ed34204c5b19bb3e07930bf1379544b47b236ec9d315eb0R179

dvordrova:
Possible ways to resolve it I see
1) either to make different settings file where commands are not templated and different test with `Commands()`
2) or keep this test that shows that using this function can lead to raw templates in data you get from settings file

Which one do you prefer?

Tarasovych:
```
b.SetCommands([]tb.Command{{Text: "test", Description: "test"}}, "", tb.CommandScopeDefault)
log.Println(b.Commands()) //[] <nil>
//why empty?
```
Is this expected behaviour?

PR #745, opened 2024-10-09
context: inherit incoming message_thread_id by default

nentenpizza:
These changes will ensure that the bot responds to incoming messages within the same topic when using context.


context.go
```diff
@@ -424,10 +424,33 @@ func (c *nativeContext) Args() []string {
 }
 
 func (c *nativeContext) Send(what interface{}, opts ...interface{}) error {
+	c.inheritContextOpts(opts)
```
demget: `inheritOpts`

demget (changes requested):

PR #776, opened 2025-06-07
api: preview options support

G74v3Dw31137:
Preview options hadn't been passed to api when embedding send options so i've implemented it.


G74v3Dw31137:
reopened: https://github.com/tucnak/telebot/pull/780

